PR #43, opened 2013-05-14
Added support for Sparrow

jovrtn:
Heya, I've found your tool extremely useful and I'd like to add a few applications. I'm not sure how you'd prefer me to go about it, but I thought I'd keep it simple and make every addition its own branch so you can selectively merge (if multiple people are trying to merge the same application, for instance). Let me know if you'd prefer it another way.

Thanks!


lra:

lra:
Sparrow is no longer active: "On July 20, 2012, the company announced that it had been acquired by [Google](https://en.wikipedia.org/wiki/Google) and was ceasing continued development of the application except for critical bug fixes."

PR #140, opened 2013-12-01
Add simulate and verbose arguments

Cottser:
I pretty much couldn't imagine using Mackup without being able to test it first using this functionality, so I decided I'd try and write it!

I'm definitely a Python newbie so please be kind and let me know if I've done anything (or everything) horribly wrong :)

I had to do some research to avoid using globals and I copped the State class from here: http://stackoverflow.com/a/747888/771643.

I didn't think it made sense to pass args to the file operation commands. Doing so would fix the `args and args.simulate` verbosity (added to fix failing tests due to no State object) but I think I'd rather have verbosity there than on the calling side. Any suggestions on improving that are definitely welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,16 @@ folder in Dropbox. This will revert your system at pre-Mackup state.
 
 Get some help, obvious...
 
+### Options
+
+`-S`, `--simulate`
+
+Simulate the Mackup operation but don't actually do anything.
+
+`-v`, `--verbose`
+
+Show verbose output for file operations.
+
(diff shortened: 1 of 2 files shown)


lra:
Cab you change `-S, --simulate` to `-n --dry-run` (more unixish) and update your PR against master? We now use docopt for the CLI.

Cottser:
Now a duplicate of https://github.com/lra/mackup/pull/140 :) closing.

lra:
cool, feel free to improve it!

Issue #37, opened 2013-05-13
Add support for Chrome


sankim:

lra:
PRed in https://github.com/lra/mackup/pull/153

PR #153, opened 2013-12-12
added support for Google Chrome

vitorgalvao:
In response to #37.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## WIP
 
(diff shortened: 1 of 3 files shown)


lra:
Depends on https://github.com/lra/mackup/issues/614

vitorgalvao:
Doing some tests on this, this isn’t a good candidate for backup. If we’re too broad directories get huge (several GBs) and if we’re too narrow we won’t backup enough.

lra:
nice, thanks for digging.

PR #160, opened 2013-12-19
VimWiki Support

amitu:
https://code.google.com/p/vimwiki/


lra:
Got it, can you update this with the new application config file format? (just merge master in)

PR #210, opened 2014-02-02
added support for readkit

vitorgalvao:
Since it’s under `Library/Containers`, this probably should wait until `mackup` can handle that.


lra:

PR #212, opened 2014-02-02
added support for sketch

vitorgalvao:


lra:

gmclelland:
Not sure it is relevant, but I thought I would mention that Sketch App is leaving the mac app store. http://blog.sketchapp.com/post/134322691555/leaving-the-mac-app-store

vitorgalvao:

Issue #195, opened 2014-01-24
Copying settings, as opposed to linking?


vitorgalvao:
I’ve been thinking about this option lately. I’ve personally had my fair share of issues caused by the way `mackup` works by default (making links to a dropbox directory), whether by having more than one computer logged on at the same time, changes to `mackup` itself, or some other reason. I’ve also seen some other open *issues* with problems that would not occur would it not be for settings being symlinked, as opposed to copied over.

How many times do users change app settings, really? Wouldn’t the extra (small) trouble of running a command when updating settings be a more than acceptable price to pay for improved stability? It would also provide some features for free, like [support for MAS apps](https://github.com/lra/mackup/issues/154).

Now, I certainly understand that would be a big change to this tool (that a lot of people use), so I’m looking for your input on this, @lra, is this something you’d be willing to consider?

mike-hearn:
I'd like to +1 this.

I use Mackup to backup my CodeKit license - this is the only thing I need backed up, with the purpose that when I `mackup restore`, it automatically adds the CodeKit license file so I don't need to re-enter my information.

The major downside here is that the license plist file also contains every other preference setting, including *window size*, so the file is getting overwritten constantly. This wastes Dropbox cycles, makes my "recent changes" pretty irrelevant, and so on. Essentially I don't care about the settings being constantly synced, I'd just like the license plist to be backed up once when I run `mackup backup`, and to be copied back to the proper place when I run `mackup restore`.

My ideal solution would be something like this:

```
[configuration]
copy = true
```

- On running `mackup backup` it just does a `cp` instead of an `ln`.
- On running `mackup restore` it does another `cp` instead of an `ln`.

I definitely think it's a common use-case for users to potentially want to back-up items (so they can be restored later), but *not* necessarily want them being persistently synced.

aristidesfl:
I also have this problem of Dropbox always syncing. Another offending application is Karabiner which causes a file update every time it is used.

lra:
> However, isn't backup/restore + uninstall = copy?

Yes

vitorgalvao:
>> However, isn't backup/restore + uninstall = copy?

> Yes

No really, no. Copying allows [possibilities that are not available with linking](https://github.com/lra/mackup/issues/154), and are thus not available with backup/restore + uninstall. Were it the same thing, this wouldn’t still be an issue.

lra:
right

lra:
Let's do it https://github.com/lra/mackup/issues/613

Issue #154, opened 2013-12-12
Support Mac App Store apps


vitorgalvao:
Mac App Store apps keep files inside `~/Library/Containers/`. These cannot be replaced with soft links (`ln -s`), but do work with hard links (`ln`) and OS X aliases. Since hard links cannot be used on directories, this would work for many apps, but not all.

My suggestion would be to add `CONTAINERS = 'Library/Containers/'` to `# Commonly used paths #`, and then when *mackup* encounters an app that references something that starts with `CONTAINERS`, use [the applescript way](http://stackoverflow.com/questions/7072208/how-do-i-create-a-macintosh-finder-alias-from-the-command-line/10067437#10067437); else, make a symbolic link as always.

dmoutal:
Also interested in the status of this.

Thanks.

lra:
Well, unless we have a working solution, there is no progress on this.
Anyone has a "working" plan?

lra:
Once https://github.com/lra/mackup/issues/613 is in, it should not be a problem

PR #213, opened 2014-02-02
added support for tweetbot

vitorgalvao:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## WIP
 
+- Added support for Tweetbot (thx @vitorgalvao)
+
 ## Mackup 0.6
 
 - Added support for custom applications
--- a/README.md
+++ b/README.md
@@ -250,6 +250,7 @@ folder and destroy the Mackup folder in Dropbox.
   - [Tower](http://www.git-tower.com/)
   - [Transmission](http://www.transmissionbt.com/)
   - [Transmit](http://panic.com/transmit/)
+  - [Tweetbot](http://tapbots.com/software/tweetbot/mac/)
   - [Twitterrific](http://twitterrific.com/)
   - [uTorrent](http://www.utorrent.com/)
   - [Ventrilo](http://www.ventrilo.com/)
--- a/mackup/applications/tweetbot.cfg
+++ b/mackup/applications/tweetbot.cfg
@@ -1,0 +1,6 @@
+[application]
+name = Tweetbot
+
+[configuration_files]
+Library/Containers/com.tapbots.TweetbotMac/Data/Library/Application Support/Tweetbot
+Library/Containers/com.tapbots.TweetbotMac/Data/Library/Preferences/com.tapbots.TweetbotMac.plist


lucaorio:
@vitorgalvao man is there any news about this?

vitorgalvao:
@lucaorio Nope, nothing I can do. Mackup [doesn’t support MAS apps](https://github.com/lra/mackup/issues/154) [because it relies on symlinking](https://github.com/lra/mackup/issues/195).

The project seems mostly abandoned feature-wise, so I somehow doubt it’ll happen. At least not without a PR.

lra:
not really abandoned, I just need people to contribute =)

vitorgalvao:
> not really abandoned, I just need people to contribute =)

Well, like I said:

> The project seems mostly abandoned feature-wise, so I somehow doubt it’ll happen. **At least not without a PR**.

So, same thing. If none of the core maintainers (you alone, I believe) is working on adding features, then the project is indeed effectively abandoned feature-wise. Wishing people contribute is very different than having said contribution.

vitorgalvao:
To be clear, I’m not criticising you not adding features. It’s your project, and as with the [multiple ssh issues](https://github.com/lra/mackup/search?q=ssh&type=Issues&utf8=%E2%9C%93), it seems to me you have it working perfectly well for your exact needs and hence have no further incentive to keep adding to or changing it, *which is totally valid*.

[Having said that](http://youtu.be/ENHVQ2gslp8), that is not what you’re saying. Issues are left to linger, and requests aren’t refused or responded to with “I won’t do it, but will accept a contribution for it”, so users rightly get the impression those things are indeed being taken into consideration and worked on, which is not the case.

lra:
I'm not against the copy feature or ssh default exclusion, but people are jumping on those while they require some other changes before, listed in the TODO.

You can't just change Mackup to optionally have a copy mode or exclude by default for ssh and not for the others in its current state. It will just make it super complex to use as some behavior will change depending on the mode or the application. It's already too complex too use IMO and it will just scare people away or generate issues with messed up configurations.

Any PR or help is welcome. Even issues to discuss the what and how to do some.

Regarding contributors, I'm def open to it, but I'm looking for code contributions here. Can't grant people just based on intention. https://github.com/lra/mackup/graphs/contributors

vitorgalvao:
Most of that was addressed in my previous comments. Quite frankly, I have little interest in continuing down this conversation since I always make a huge effort to be respectful in pointing out what I think is wrong here but am always faced with what feels like a defensive take and answers to points I never made.

My take on it is (I believe) very clear: it is perfectly OK for you to not work on the project and everything that is asked, but you don’t say it. Issues are left open, *including* ***numerous*** *duplicates* without reply or acknowledgement for a long time (*or ever*). I’m certain there are capable users that would take on the job of submitting PRs for what is needed, *but if you don’t participate on the discussion of what you believe should be included* ***and how to implement it*** *why would anyone else*?

PRs won’t happen simply because there’s a TODO. Contributors need to feel their ideas will be at the very least considered (even if immediately refused) and that there is a point to spending time on them. If you don’t participate, they never do, and move on.

lra:
I'm sorry if this is what is perceived. You make a good point managing the PRs and issues: it also warrants some attention.

Some help here would be appreciated. Also to point out those that lack some feedback, direction or approval to get going. BTW the existing issue/PR organization can be totally revamped.

I can add you if you are interested as you've been a long time help.

vitorgalvao:
I appreciate the offer, but I honestly do not think I’d be a good candidate for that as I stopped using mackup, since syncing instead of copying broke too many things for me. For that reason I don’t check back as often, and you’d be back at square one.

I’m still up for helping with unifying the configs into a common repo/format, as we’ve talked about other times, though.

I do however recommend that when you engage with the issues, keep and eye out for a user that frequently participates, and “promote” that one to that position. Many times giving people the extra responsibility makes them work even more to earn it (in homebrew-cask we have examples).

lra:
As you want, but know that I'm not asking you to be active. If you want to spend some time on it when you feel like it that'd be fine. It's not like we are limited in the number of contributors =)

lra:

vitorgalvao:

PR #221, opened 2014-02-27
added support for fontcase

vitorgalvao:
No long under development, but still one of the best at what it does.

Since it’s under `Library/Containers`, this probably should wait until `mackup` adds support for that.


lra:

vitorgalvao:
Actually, Fontcase is dead. It has been discontinued for a while.

lra:
Cool thanks

PR #236, opened 2014-03-13
Configuration Generation

Timidger:
If ~/.mackup.cfg does not exist, Mackup asks the user if he wants to make one. It then prompts them to answer some questions so that it can be generated. Also generates custom apps in the ~/.mackup folder if the user wants to.

### DIFF
--- a/mackup/config.py
+++ b/mackup/config.py
@@ -24,7 +24,50 @@ try:
 except ImportError:
     import ConfigParser as configparser
 
-import appsdb
+
+def make_config_file(storage_type=ENGINE_DROPBOX, path="",
+                     directory=MACKUP_BACKUP_PATH, whitelist=None,
+                     blacklist=None):
+    """
+    Writes a configuration file (~/.mackup.cfg) using the parameters. If no
+    options are specified, a default configuration file is made.
+
+    Args:
+        (str): The storage_type to be written (dropbox, google_drive,
+        file_system)
+        (str): The path to use if storage_type was file_system
+        (str): The custom directory name that will be used
+        (list): The list of applications that will be synced
+        (list): The list of applications that will not be synced
+    """
+    # Get the names of the apps to check the whitelist and blacklist
+    if not whitelist:
+        whitelist = []
+    if not blacklist:
+        blacklist = []
+    # Make sure the storage_type is correct
+    assert storage_type in (ENGINE_DROPBOX, ENGINE_FS,
+                            ENGINE_GDRIVE)
+    # If the path is specified, make sure it exists
+    if path:
+        assert os.path.exists(path), (
+            "The path {} does not exist!".format(path))
+    # Write the configuration file in the home directory
+    configuration = ["[storage]",
+                     "engine = {}".format(storage_type),
+                     ("path = " + (path or "")) if storage_type else "",
+                     "directory = {}".format(directory),
+
+                     "[applications_to_sync]",
+                     "\n".join(whitelist),
+
+                     "[applications_to_ignore]",
+                     "\n".join(blacklist)
+                     ]
+    config_path = os.path.join(os.environ['HOME'],
+                               MACKUP_CONFIG_FILE)
+    with open(config_path, "w") as config_file:
+        config_file.writelines(line + "\n" for line in configuration)
 
 
 class Config(object):
@@ -52,7 +95,6 @@ class Config(object):
 
         # Get the path where the Mackup folder is
         self._path = self._parse_path()
-        print self._path
 
         # Get the directory replacing 'Mackup', if any
         self._directory = self._parse_directory()
@@ -180,10 +222,8 @@ class Config(object):
             str
         """
         if self._parser.has_option('storage', 'engine'):
-            print "Yes!"
             engine = str(self._parser.get('storage', 'engine'))
         else:
-            print "No"
             engine = ENGINE_DROPBOX
 
         assert isinstance(engine, str)
--- a/mackup/config.py
+++ b/mackup/config.py
@@ -53,22 +53,31 @@ def make_config_file(storage_type=ENGINE_DROPBOX, path="",
     if storage_type == ENGINE_FS and path:
         assert os.path.exists(path), (
             "The path {} does not exist!".format(path))
+    else:
+        path = ""
     # Write the configuration file in the home directory
-    configuration = ["[storage]",
-                     "engine = {}".format(storage_type),
-                     ("path = " + (path or "")) if storage_type else "",
-                     "directory = {}".format(directory),
+    # This writes it backwards, so we have to specify it backwards
+    config = configparser.ConfigParser()
 
-                     "[applications_to_sync]",
-                     "\n".join(whitelist),
+    # Applications to NOT sync section
+    config.add_section('applications_to_ignore')
+    for app in blacklist:
+        config.set('applications_to_ignore', app, "")
 
-                     "[applications_to_ignore]",
-                     "\n".join(blacklist)
-                     ]
-    config_path = os.path.join(os.environ['HOME'],
-                               MACKUP_CONFIG_FILE)
+    # Applications to sync section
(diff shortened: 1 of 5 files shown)


lra:
After giving it more thoughts, I made up my mind of having mackup to syncing anything by default. See https://github.com/lra/mackup/issues/614

But this is still super useful, except that it should be triggered by the user.

Can you make it an explicit action (e.g. `mackup configure`) instead of checking if the config file exist?

Timidger:
Being explicit here is the better option, especially since a user might not want to answer a bunch of questions after downloading some software for the first time.  I'll get to work switching this over to be an explicit command.

Timidger:
I switched it over to a command (`mackup configure`), and also updated the storage options to include the new ones since the patch was first written.

mackup/config.py
```diff
@@ -25,6 +25,51 @@
...
+        (list): The list of applications that will be synced
+        (list): The list of applications that will not be synced
+    """
+    # Get the names of the apps to check the whitelist and blacklist
+    if not whitelist:
+        whitelist = []
+    if not blacklist:
+        blacklist = []
+    # Make sure the storage_type is correct
+    assert storage_type in (ENGINE_DROPBOX, ENGINE_FS,
```
lra: we have a lot more now

mackup/config.py
```diff
@@ -25,6 +25,51 @@
...
+                            ENGINE_GDRIVE)
+    # If the path is specified, make sure it exists
+    if path:
```
lra: only allow this if the ENGINE_FS is used.

mackup/config.py
```diff
@@ -25,6 +25,51 @@
...
+        assert os.path.exists(path), (
+            "The path {} does not exist!".format(path))
+    # Write the configuration file in the home directory
+    configuration = ["[storage]",
```
lra: Can you use Python's ConfigParser to create this file content? Way easier and solid than trying to build it with strings.

mackup/config.py
```diff
@@ -274,6 +319,69 @@ def _parse_apps_to_sync(self):
 
         return apps_to_sync
 
+    @staticmethod
+    def make_config_file():
+        storage_type = choose("What method would you like to use to backup "
+                              "your configuration files?",
+                              ["dropbox", "google_drive", "file_system"])[0]
```
lra: we have more

mackup/utils.py
```diff
@@ -9,6 +9,50 @@
...
+        default(str): If given and in the options list, the default is used
+        if the user does not give an option
+
+    Returns:
+        (list): The option(s) they chose from the list
+    """
+    default_str = ("(default: {})".format(default) if default in options
+                   else "")
+    while True:
+        answer = raw_input("{} {} <{}>\n".format(question, default_str,
```
lra: if default_str is empty, you'll have 2 spaces, which look bad.

mackup/config.py
```diff
@@ -274,6 +319,69 @@ def _parse_apps_to_sync(self):
...
+
+        directory = "Mackup"
+        if confirm("Would you like to customize the directory in which "
+                         "Mackup stores your files?"):
+            directory = raw_input("Directory name: ")
+
+        whitelist = ""
+        if not confirm("Would you like to sync all applications or "
+                              "specify from a list of supported software?"):
+            whitelist = choose("Choose any of the follwing "
```
lra: following

mackup/config.py
```diff
@@ -274,6 +319,69 @@ def _parse_apps_to_sync(self):
...
+            "deliminated by a space): ",
+            appsdb.ApplicationsDatabase().get_pretty_app_names(), True)
+
+        blacklist = ""
+        if confirm("Would you like to specify any applications you would "
+                        "NOT like to sync?"):
+            blacklist = choose("Choose any of the follwing "
```
lra: following

mackup/config.py
```diff
@@ -274,6 +319,69 @@ def _parse_apps_to_sync(self):
...
+            "deliminated by a space): ",
+            appsdb.ApplicationsDatabase().get_pretty_app_names(), True)
+
+        custom_apps = {} # {app_name: [config files]}
+        if confirm("Now, would you like to add any applications not "
+                         "currently supported by Mackup?"):
+            print "Enter nothing to indicate your done"
```
lra: that you are done

mackup/config.py
```diff
@@ -274,6 +319,69 @@ def _parse_apps_to_sync(self):
...
+            while True:
+                name = raw_input("Application name: ")
+                if not name:
+                    break
+                config_files = raw_input("Please enter the names of all the "
+                "configuration files, delemenated by a space").split()
```
lra: delimited

mackup/config.py
```diff
@@ -274,6 +319,69 @@ def _parse_apps_to_sync(self):
...
+
+        custom_apps = {} # {app_name: [config files]}
+        if confirm("Now, would you like to add any applications not "
+                         "currently supported by Mackup?"):
+            print "Enter nothing to indicate your done"
+            while True:
+                name = raw_input("Application name: ")
+                if not name:
+                    break
+                config_files = raw_input("Please enter the names of all the "
```
lra: This is not gonna work, as paths can have spaces in them.

mackup/config.py
```diff
@@ -274,6 +319,69 @@ def _parse_apps_to_sync(self):
...
+                "configuration files, delemenated by a space").split()
+                custom_apps.update({name: config_files})
+        with open(os.path.join(os.path.expanduser("~"),
+                   MACKUP_CONFIG_FILE), "w") as config_file:
+            config_file.write("[storage]\n")
```
lra: Same here, you should use Python's ConfigParser

mackup/utils.py
```diff
@@ -407,3 +451,122 @@ def can_file_be_synced_on_current_platform(path):
             can_be_synced = False
 
     return can_be_synced
+
+
+def get_storage_config():
+    """
+    Asks the user to enter storage information written that will later be
```
lra: double written

mackup/utils.py
```diff
@@ -407,3 +451,122 @@ def can_file_be_synced_on_current_platform(path):
...
+            exists = os.path.exists(
+                ("" if path.startswith(os.sep) or path.startswith("./")
+                else os.environ['HOME']) + path)
+            if exists:
+                break
+            else:
+                print "The path {} does not exist!".format(exists)
+
+    while True:
+        directory = (raw_input("What would directory name would you like to "
```
lra: Which instead of What would

mackup/utils.py
```diff
@@ -407,3 +451,122 @@ def can_file_be_synced_on_current_platform(path):
...
+        blacklist = []
+    # Make sure the storage_type is correct
+    assert storage_type in (constants.ENGINE_DROPBOX, constants.ENGINE_FS,
+                            constants.ENGINE_GDRIVE)
+    # If the path is specified, make sure it exists
+    if path:
+        assert os.path.exists(path), (
+            "The path {} does not exist!".format(path))
+    # Write the configuration file in the home directory
+    configuration = ["[storage]",
```
lra: Same, use ConfigParser

Timidger:
Switched over the configuration file generation from the hacky string-based one to ConfigParser's built-in mechanism for it. I had to build it backwards though (the docs says it writes it backwards for some reason). Tested it myself as well, it generates perfectly.

Timidger:
Updated the files. I deleted that config generator function in `utils.py`, since I moved it to `config.py` (all of its helper functions are still in `utils.py` though).

paour:
Too bad this PR is languishing, it fixes a real pain point in mackup…

sethherr:
:+1: I didn't know .mackup.cfg existed until looking for .gitignore_global, having documentation of it and a tool to use it (my first .mackup.cfg attempt failed because I didn't include headers) would be great.

lra:
@sethherr PR welcome ;)

PR #242, opened 2014-03-22
Added the ability to "toggle" apps

Timidger:
Added the enable and disable commands. When supplied with a list of apps, the enable and disable commands edits the configuration file for mackup. If using the enable command, then it adds the app names to the "applications_to_sync" section. If using the disable command, then it adds the app names to the "applications_to_ignore" section. Before writing to the configuration file, the app names are removed from BOTH sections so that neither conflicts or duplicates are written to the configuration file.


lra:
There multiple changes in this PR.
Can you isolate different PRs for each newly added application?

As of the enable/disable feature, as you mentioned, it's not possible yet, see https://github.com/lra/mackup/issues/608 for the dependancies.

Once the dependancies are in, we can shape this PR to solve https://github.com/lra/mackup/issues/608

PR #338, opened 2014-10-04
Update sublime-text-3.cfg

babygau:
- Backup `Packages` and `Installed Packages` folders which allow user to fully restore not only configuration but also plugins


lra:
What about saving the whole `Library/Application Support/Sublime Text 3` then?

thibaultCha:
:+1: to this

lra:
But that goes against what's being said at https://packagecontrol.io/docs/syncing

thibaultCha:
@ira Indeed, I missed #169 which was mentioning that.

PR #344, opened 2014-10-05
Support Copy in the sync engine.

exiva:
Added support for [Copy](https://copy.com) in the sync engine.


lra:
That's awesome, looks good

Issue #336, opened 2014-10-03
.ssh directory copied by default


lambdahands:
Loving this project so far, but I noticed my `~/.ssh` directory including my `id_rsa` were copied by default to my Dropbox account. This is, well, uncomfortable. It may be best to disable backing up keys to a cloud service by default.

llinder:
Optionally encrypting files might be a good approach?  This way storing these files in dropbox is less worrisome.

lra:
Dupe of https://github.com/lra/mackup/issues/109

PR #295, opened 2014-07-13
Refine ssh config: no need to backup keys

nkcfan:
It is safer to prevent any the credential keys shared or backed up. Normally if computer changed, the user should generate new keys instead of reusing existing ones.

PR #352, opened 2014-10-06
Only backs up sensible SSH settings.

ISNIT0:
https://github.com/lra/mackup/issues/336

### DIFF
--- a/mackup/applications/ssh.cfg
+++ b/mackup/applications/ssh.cfg
@@ -1,5 +1,0 @@
-[application]
-name = SSH
-
-[configuration_files]
-.ssh
--- a/mackup/applications/ssh.cfg
+++ b/mackup/applications/ssh.cfg
@@ -1,0 +1,7 @@
+[application]
+name = SSH
+
+[configuration_files]
+~/.ssh/config
+~/.ssh/known_hosts
+~/.ssh/authorized_keye


RickWong:
This should be a no-brainer. You don't mess with someone's privates unless they consent to it.

lra:
@RickWong Any config synced with mackup is private, that's the essence of it and I made it to sync my ssh keys in the 1st place. The problem is that it's now used by other people =)

RickWong:
Difference is that ssh keys are **keys** not config (.ssh/config). For some users syncing ssh keys can indeed be convenient. The question here is: Should mackup sync **keys** by opt-out (default) or by opt-in?

Personally I found it rather unpleasant to find that Dropbox* obtained my ssh key, just because I used `ls -lah` to view the backup dir. Only after that discovery did I find this issue on Github. Had to go replace my ssh keypairs everywhere.

* I trust Dropbox, but not this much.

mackup/applications/ssh.cfg
```diff
@@ -2,4 +2,6 @@
 name = SSH
 
 [configuration_files]
-.ssh
+~/.ssh/config
+~/.ssh/known_hosts
+~/.ssh/authorized_keye
```
sqren: It seems like this should be 
```
[configuration_files]
.ssh/config
.ssh/known_hosts
.ssh/authorized_keys
```

lra:
Dupe of https://github.com/lra/mackup/pull/295

PR #370, opened 2014-10-15
Added support for npm

zheng1:


mackup/applications/npm.cfg
```diff
@@ -0,0 +1,6 @@
+[application]
+name = npm
+
+[configuration_files]
+.npmrc
+
```
lra: too many newline ;) will remove it

lra:
You also need to add it to the README, will do.

PR #374, opened 2014-10-18
Use lists instead of sets for all config files

Gyran:
Because sets are Unordered, it's not always sure that custom apps overrides the default app config


lra:
I see, but I'd rather fix the underlying problem. We should not rely on order for a list of files that should be unique in the first place. The fix should be that in `get_config_files()`, for each application, we should nott add the stock application config if there is a custom.

I will work on a fix.

PR #376, opened 2014-10-18
Add Ubersicht

kfinlay:


lra:
@jannae we can try the default ones, and see how the community will improve this

PR #389, opened 2014-11-09
Fix application config file precedence

jalaziz:
The docs mention that a custom application config file would overwrite a stock config file. However, using a `set` won't guarantee this behavior. Using a `list` would force the custom config files to be parsed after the stock ones.


lra:
Dupe of https://github.com/lra/mackup/pull/374

PR #391, opened 2014-11-22
Added iStat Menus 5 support

jverdeyen:


swrobel:
This working for everyone else? iStat Menus 5.11 seems to keep overwriting `Library/Preferences/com.bjango.istatmenus5.extras.plist` with a non-symlinked default version

jverdeyen:
Feels like I'm having the same issue with 5.11.

lra:
If this is broken, a PR to remove support is welcome

guillaumealgis:
Same here, reproduced the problem reliably multiple times with iStatMenus 5.20 (682).
I contacted bjango to let them know of the issue. If they don't respond or don't address the issue, I'll submit a PR removing the app :'(

guillaumealgis:
Well, they did not respond to my ticket but they fixed it anyway, the symlink is no longer overwritten anymore on 5.2 🎉

PR #412, opened 2014-12-10
OsX Stickies

dbbz:
Added support for the native lightweight notes application on OsX.


zanderzhng:
This actually doesn't work as excepted. Stickies App will overwrite the ~/Library/StickiesDatabase file everytime it saves,  even though it might be a symbolic link.

Details [here](http://jessenma.tumblr.com/post/15431886810/sync-stickies-with-dropbox-on-your-mac)

lra:
@zanderzhng so should we just remove complete support for it?

PR #415, opened 2014-12-12
fix st3 to only sync essential files, excluding cache, session, etc

hiyer:
Sublime Text 3 configuration was syncing the entire config folder, which includes files that are essentially "local", like session information, cache, backups, etc. This PR fixes that on Linux.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - Support for Scroll Reverser (via @ponychicken)
 - Support for JuliaLang (via @danielsuo)
 - Support for OsX Stickies (via @djabbz)
+- Sublime Text 3 now only syncs essential files, excluding session and cache folders
 
 ## Mackup 0.7.4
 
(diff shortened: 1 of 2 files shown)


jnv:
This is still too much, [the recommended settings](https://packagecontrol.io/docs/syncing) are to sync just `Packages/User` folder – this is a current Mackup's behaviour on Mac OS X too.

lra:
@jnv Nice link, I'll fix it

lra:
u did, thanks

hiyer:
@jnv Packages/User only contains the settings for the installed packages, and not the packages themselves. If you want to sync the packages as well, you need Installed Packages and Packages folders. This was my intention when I added these two. Still, I suppose it makes sense to use the recommended settings - if you're syncing across different OSes, the same packages may not work.

lra:
@hiyer That's not what the page says:
> If this file is copied to another machine, the next time Sublime Text is started, Package Control will install the correct version of any missing packages.
https://packagecontrol.io/docs/syncing

hiyer:
Ah ok, makes sense. Thanks for the correction!

PR #434, opened 2015-01-03
Add support for i3-wm tiling manager.

akash0x53:


mackup/applications/i3wm.cfg
```diff
@@ -0,0 +1,8 @@
+[application]
+name=i3wm
+
+[configuration_files]
+.i3/config
+.i3status.conf
+.xsession
+
```
vitorgalvao: This empty line shouldn’t be present.

PR #435, opened 2015-01-03
Add support for i3-wm tiling manager.

akash0x53:


mackup/applications/i3wm.cfg
```diff
@@ -0,0 +1,7 @@
+[application]
+name=i3wm
+
+[configuration_files]
+.i3/config
+.i3status.conf
+.xsession
```
lra: .xsession is not i3wm specific, it's more a X11 thing right?

akash0x53: Yeah, you're right, its not i3 specific, however its good to keep with i3wm configs.

lra: We can't have configs leak from one program to another, or every program will have all the configs. It should be in the X11 config.

akash0x53:
Okay, removed .xsession.

lra:
And you need to add yourself to the changelog ;)
https://github.com/lra/mackup/blob/master/CONTRIBUTING.md

PR #441, opened 2015-01-14
Update sublime-text-3.cfg

GillesDebunne:
- Backup `Packages` and `Installed Packages` folders which allow user to fully restore not only configuration but also plugins

- Remove `.config/sublime-text-3` references since they are not listed in the documentation
  (http://www.sublimetext.com/docs/3/settings.html)

### DIFF
--- a/mackup/applications/sublime-text-3.cfg
+++ b/mackup/applications/sublime-text-3.cfg
@@ -2,6 +2,5 @@
 name = Sublime Text 3
 
 [configuration_files]
-Library/Application Support/Sublime Text 3/Packages/User
-.config/sublime-text-3/Installed Packages
-.config/sublime-text-3/Packages
+Library/Application Support/Sublime Text 3/Packages
+Library/Application Support/Sublime Text 3/Installed Packages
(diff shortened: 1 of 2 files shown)


GillesDebunne:
This fixes the merge conflicts in #338, and removes `.config` files from the list.

lra:
`.config/sublime-text-3` are the config paths used on GNU/Linux. See https://www.sublimetext.com/docs/3/revert.html

mackup/applications/sublime-text-3.cfg
```diff
@@ -2,6 +2,7 @@
 name = Sublime Text 3
 
 [configuration_files]
-Library/Application Support/Sublime Text 3/Packages/User
```
hiyer: According to this: https://packagecontrol.io/docs/syncing, we should sync the Packages/User folder and not Packages and Installed Packages. I think the earlier configuration was correct.

lra: That's what is live https://github.com/lra/mackup/blob/master/mackup/applications/sublime-text-3.cfg
So I should close this PR?

hiyer: Yes, IMO this change is unnecessary. I have verified that the current configuration works (had to set up a new laptop just yesterday) so I don't see any reason to change it. Perhaps the source of confusion - and why some folks think this is not working - is that the Sublime Package Control package needs to be installed manually. Once that is done, it reads some configuration files and pulls the other packages automatically. If we sync Packages and Installed Packages, probably that manual step would not be required, but given the author (of Package Control) advises against it I'm not sure if mackup should work that way.

lra: ok thanks

PR #322, opened 2014-08-21
Added DefaultKeyBinding.dict

kfinlay:
https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/EventOverview/TextDefaultsBindings/TextDefaultsBindings.html

PR #452, opened 2015-01-31
Defaultkeybinding

adamclerk:
I know this isn't exactly an application. If you would like me to include this in an existing app's config please let me know.


adamclerk:
I noticed this PR here...
https://github.com/lra/mackup/pull/322

please advise how you would like to proceed.

PR #454, opened 2015-02-04
100 % test coverage for config tests

Timidger:
Added tests for when there is an old config (which throws a `SystemExit` exception) and for when there is no config file present (which should do the same thing as a default configuration file)

Proof: 
test-coverage

### DIFF
--- a/tests/config_tests.py
+++ b/tests/config_tests.py
@@ -184,3 +184,6 @@ class TestConfig(unittest.TestCase):
                                         'sublime-text-3',
                                         'x11',
                                         'vim'])
+
+    def test_config_old_config(self):
+        self.assertRaises(SystemExit, Config, 'mackup-old-config.cfg')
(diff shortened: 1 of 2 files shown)


tests/config_tests.py
```diff
@@ -13,6 +13,12 @@ def setUp(self):
         realpath = os.path.dirname(os.path.realpath(__file__))
         os.environ['HOME'] = os.path.join(realpath, 'fixtures')
 
+    def test_config_no_config(self):
+        cfg = Config()
```
lra: isn't it overiden by `cfg = Config('mackup-empty.cfg')` in `test_config_empty()`? Don't worry to copy paste in unit tests, that's fine.

Timidger: Oops, you're right. Good catch

lra:
Cool thx

PR #455, opened 2015-02-04
Update bash.cfg

dmarg:
Update bash.cfg to include .bash_prompt and .functions (for bash functions)

### DIFF
--- a/mackup/applications/bash.cfg
+++ b/mackup/applications/bash.cfg
@@ -10,3 +10,5 @@ name = Bash
 .profile
 .bash_profile
 .inputrc
+.bash_prompt
+.functions


lra:
Can you point to an official doc mentioning those files? If it's just for you, it should be a user conf.
Also can you follow https://github.com/lra/mackup/blob/master/CONTRIBUTING.md
Thanks =)

dmarg:
I have not seen any official documentation on this but I have seen a lot of people around use .bash_prompt and .functions as well as others like .exports. I have gone ahead and created a user conf so I guess we can cancel this request. What about creating a new "app" called bash-extras and we .bash_prompt, .functions .exports?

lra:
The thing is that it's the start to have `.*` synced, and we need a clear and explicit frontier ;)
As of now, it as been: if it's not official, it should not be committed. If you have something better in mind, do a proposal ;)

dmarg:
Alright sounds good. I will think on it but for now I will close this request.  Thanks and thanks for making this. I just found it and it is pretty sweet!

lra:
cool thx

PR #466, opened 2015-02-10
Adding support for Photoshop CC

ponychicken:
This is split up into two cfg files: Adobe releases about one major version of Photoshop per year. The major versions each use a new set of Preference folders. I think it's best to have one cfg per version, but it would be also possible to put them all in one. Thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## WIP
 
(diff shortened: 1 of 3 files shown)


lra:
It's fine keeping all them in one config. Did you try the config?

ponychicken:
Ok, I put it all into one file. It synced correctly between two machines and photoshop didn't seem to mind.

PR #470, opened 2015-02-16
Added support for Mac OS X Dock.app

fcvarela:


lra:
IMO it's a bad idea to sync preferences that depend on what is installed.
E.g. If I have an application icon in the dock on 1 computer, and sync the dock preferences on another computer where this application is not installed, what will happen?

Not a fan of those in general, but maybe I'm missing something?

fcvarela:
I actually use this to keep the dock in sync on several machines I use on which I want the exact same setup. I don't see any other reason why anyone would want to keep the dock's prefs in sync. Totally see your point, but think it applies to cases where users would choose not to sync it.

lra:
Ok, it might be a legit thing to sync yeah. I'd welcome another opinion on this =)

fcvarela:
+1, would love alternatives. OS X actually caches defaults so unless you
'defaults read plistfile' and 'defaults read com.apple.<something>' it
simply won't work.

On Mon, Feb 16, 2015 at 10:11 PM, Laurent Raufaste <notifications@github.com

lra:
As it's only gonna work in certain use case, let's wait for https://github.com/lra/mackup/issues/614 1st before adding this.

fcvarela:
Not much point in keeping this open

PR #477, opened 2015-02-28
sublime-text-3: Sync only User folder on Linux

jnv:
As I have commented in https://github.com/lra/mackup/pull/415#issuecomment-76547492, only `Packages/User` folder [should be synced](https://packagecontrol.io/docs/syncing#dropbox-linux), which is the current behaviour on Mac OS X too.


lra:
Also did https://github.com/lra/mackup/commit/9ca6e273bbd30c5569e3a672f8413333ac8b6bc1

jnv:
Many thanks!

PR #487, opened 2015-03-11
Custom themes and plugins support

joaoponceleao:
When using antigen or other zsh plugin managers, custom themes and plugins are found in the `.zsh/themes` and `.zsh/plugins` dir. If not backed up these are lost.
This takes care of that.
Feel free to chime in if you feel this should be done differently.

### DIFF
--- a/mackup/applications/zsh.cfg
+++ b/mackup/applications/zsh.cfg
@@ -7,3 +7,5 @@ name = Zsh
 .zshrc
 .zlogin
 .zlogout
+.zsh/plugins
+.zsh/themes


lra:
Are there official mentions of those paths in the zsh doc?

joaoponceleao:
No there isn't. Damn.
I realise now I must have made that path up to mimick the oh-my-zsh folder.
Antigen will manage zsh plugins and themes itself (wherever they may come from: ohmyzsh, prezto, github, etc - hence I felt custom stuff did not belong in a ohmyzsh folder), but there needs to be a consensus of where this local custom stuff goes. This is probably an issue for antigen rather than mackup in that case.
Sorry for that. I'll close the issue.

lra:
Cool thanks.

PR #492, opened 2015-03-16
Added -v and --version flag to the command line interface for Mackup

Timidger:
I found it annoying having to use `mackup -h` each time I needed to check the version number, and then having to hunt for it amongst all the other help text. So I just whipped this up so it was easier to compare versions quickly and efficiently.


lra:
Me too ;)

Issue #283, opened 2014-06-26
"Are you sure that your want to replace it?" <Yes|No>


ryanburnett:
When I `mackup backup` I receive this prompt for most files:

> A file named /Users/Ryan/Dropbox/Mackup/Library/Preferences/com.irradiatedsoftware.SizeUp.plist already exists in the backup.
Are you sure that your want to replace it ? \<Yes|No>

...I had to type `Yes` about 10 times. First world problems, I know ;)

* Anyway to add an `--overwrite` flag or similar to avoid having to type `Yes` for each updated file?
* *Spellcheck police* `your` should be `you` :)

Thanks for making mackup!

Anahkiasen:
:+1: for a `--force` flag, always receiving prompts every time I backup from Popcorn Time and SourceTree

drawcard:
I'd love this feature, I currently have to copy and paste the word "Yes" into the terminal repeatedly to stop these questions from coming up repeatedly.

kiliankoe:
For things that require you to do this you can just pipe the output of yes into it ;)

`yes | mackup backup`

drawcard:
Ah. Well that makes a lot of sense! :)

lra:
Fixed by https://github.com/lra/mackup/pull/493

Issue #68, opened 2013-05-20
'Yes' to all option on restore


TimCorcoran:
A good idea (IMO) would be to have a Yes to all option for the 'Overwrite blah blah' on mackup restore. Currently you have to type 'Yes' (full word, no abbreviations) to each conflict, a 'Yes to All' option would save some time an enhance user experience. It's probably a good idea though that if it is implemented, it gives you a list of all the conflicts at once before you can 'Yes to All' without knowing what 'All' is (if you get my drift.

drawcard:
A rather dodgy way I've just discovered, is to copy the word "Yes" to your clipboard and use Command-V to paste it into the Terminal and hit multiple times to flip through the repetitive questions :/

lra:
PR up for this https://github.com/lra/mackup/pull/493

PR #493, opened 2015-03-17
Added the -f and --force argument to the cmdline parser

Timidger:
I found it annoying to type out "y" or "Yes" each time I re-backed up mackup while I was fiddling around with it, so I added a `-f` and `--force` feature that basically answers "yes" to every question mackup asks. This should be enough to fix issues #283 and #68.

### DIFF
--- a/mackup/main.py
+++ b/mackup/main.py
@@ -47,6 +47,10 @@ def main():
     mckp = Mackup()
     app_db = ApplicationsDatabase()
 
+    # If we want to answer mackup with "yes" for each question
+    if args['force']:
+        utils.FORCE_YES = True
+
     if args['backup']:
         # Check the env where the command is being run
         mckp.check_for_usable_backup_env()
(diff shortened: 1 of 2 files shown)


mackup/main.py
```diff
@@ -36,6 +36,10 @@ def main():
     mckp = Mackup()
     app_db = ApplicationsDatabase()
 
+    # If we want to answer mackup with "yes" for each question
+    if args.force:
+        utils.raw_input = lambda *args: "Yes"
```
lra: I find this a little heavy =) Can you just check for the force argument in the utils.confirm function instead?

lra:
Since your PR, we switched to docopt to handle the CLI. Mind updating your PR?

Timidger:
I've updated the pull request to use Docopt. I added a simple flag for force confirmation, so now it isn't as hacky.

lra:
Cool, but for docopt to accept the --force option, it needs to be added here https://github.com/lra/mackup/blob/master/mackup/main.py#L14

Can you update your PR?

Timidger:
There we go, added the force description where docopt needs it to be

PR #498, opened 2015-03-30
Add support for common global gitignores.

musophob:
Covers both `~/.gitignore_global` and `~/.gitignore`.

The Github instructions for setting a global ignore propose using
`~/.gitignore_global`. Most probably have this.
https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

Another large group probably have simply `~/.gitignore`.
https://robots.thoughtbot.com/global-gitignore

### DIFF
--- a/mackup/applications/git.cfg
+++ b/mackup/applications/git.cfg
@@ -4,3 +4,5 @@ name = Git
 [configuration_files]
 .gitconfig
 .config/git/ignore
+.gitignore_global
+.gitignore


lra:
Again, mackup can't support all the files for all the behavior in the world =)
Where are those files officially mentioned by the git doc our source code?
If not this is more a `.mackup.cfg` thing, unless a lot of people think this is a defacto standard.

musophob:
You're right, I may be a bit overzealous about mackup. My reasoning is that

1. Git is extremely popular and common amongst the kind of user that would setup mackup (more so than say iterm2, adobe, etc
2. There's already a git cfg dir in mackup, might as well make it more complete. I think a global git ignore is a very common setting. Often when I switch machines it leads to accidental commits of garbage files because I don't have all the random junk that was incrementally appended to my global gitignore in the past.
3. Googling for "global gitignore" almost certainly returns [this github article](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) (w and w/out personalized search results), which demos `~/.gitignore_global`. The majority of global gitignore users most likely reference that article or simply go with `~/.gitignore` as is the convention for project-based ignores according to the git docs.

gwerbin:
I came here because I Google'd this exact situation.

I think this can be easily rectified by more prominently highlighting the "roll your own" feature in the main Readme.

lra:
> I think this can be easily rectified by more prominently highlighting the "roll your own" feature in the main Readme.

Feel free to PR =)

sethherr:
:+1: for it being a defacto standard that many people have.

I'm now adding it to my mackup config, but it seems like it's maybe something that should just be in there.

PR #503, opened 2015-04-11
CotEditor support 

usami-k:
CotEditor -Text Editor for OS X
http://coteditor.com/


usami-k:
Re-send pull-request (#533)

PR #513, opened 2015-04-15
Add support for .git-completion.bash

hannupekka:
Adds support to backup https://github.com/git/git/blob/master/contrib/completion/git-completion.bash

### DIFF
--- a/mackup/applications/git.cfg
+++ b/mackup/applications/git.cfg
@@ -4,3 +4,4 @@ name = Git
 [configuration_files]
 .gitconfig
 .config/git/ignore
+.git-completion.bash


mackup/applications/git.cfg
```diff
@@ -4,3 +4,4 @@ name = Git
 [configuration_files]
 .gitconfig
 .config/git/ignore
+.git-completion.bash
```
lra: newline?

lra:
I don't see any place where it's stated that `~/.git-completion.bash` is the official place for this file, looks like it's pretty user specific.

hannupekka:
Oh, my bad. Although for example this http://git-scm.com/book/en/v1/Git-Basics-Tips-and-Tricks and https://github.com/git/git/blob/master/contrib/completion/git-completion.bash suggest saving it to that location, but it is indeed user specific.

PR #536, opened 2015-04-27
Added support for xbindkeys

ScottAmes:
Added support for [xbindkeys](http://www.nongnu.org/xbindkeys/) - "a program that allows you to launch shell commands with your keyboard or your mouse under X Window."

Thanks for such a great tool!


mackup/applications/xbindkeys.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = xbindkeys
+
+[configuration_files]
+.xbindkeysrc
```
lra: can you finish this line with a newline char?

ScottAmes: Absolutely, sorry about that.

PR #537, opened 2015-04-28
Added support for Spacemacs

x-ji:
Spacemacs https://github.com/syl20bnr/spacemacs is an Emacs advanced Kit focused on Evil. It modifies `.emacs.d`(which is already backed up in Mackup) but also creates its own dotfile called `.spacemacs` in home folder.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Added support for Royal TSX (via @amatos)
 - Added support for CotEditor (via @usami-k)
 - Added support for xbindkeys (via @scottames)
+- Added support for Spacemacs (via @x-ji)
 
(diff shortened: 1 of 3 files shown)


lra:
Can you add `.emacs.d` if it's a dependancy?

x-ji:
@lra Sure. I thought `emacs.cfg` already specified that though. That wouldn't be an issue?

lra:
nope it should not be

lra:
`.emacs` too?

lra:
mmmh maybe you are right we should not merge them and rely on the user to explicitely activate support for emacs.
Can you remove it? =)

x-ji:
Done

PR #538, opened 2015-04-29
Add Visual Studio Code support

luisdemanuel:


mackup/applications/vscode.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = vscode
```
lra: The name should be the full real name, it's for display purpose

mackup/applications/vscode.cfg
```diff
@@ -0,0 +1,5 @@
...
+
+[configuration_files]
+Library/Application Support/Code/User
```
lra: nothing big in this thing?

lra:
Can you add a line in the changelog? https://github.com/lra/mackup/blob/master/CONTRIBUTING.md

mackup/applications/vscode.cfg
luisdemanuel: No, only a keybindings and a settings JSON files

PR #539, opened 2015-05-10
don't break textmate's ruby

samsonjs:
TextMate 2's ruby is broken when mackup symlinks it. We can be a bit more selective about what is synced as there are only a few files in Library/Application Support/TextMate. The other files probably don't need to by synced as they are system dependent. Clipboard history, last window configuration. But maybe they could be. The only real problem is syncing the Ruby folder. If there's a way to exclude a file maybe we could use that instead.

### DIFF
--- a/mackup/applications/textmate.cfg
+++ b/mackup/applications/textmate.cfg
@@ -2,6 +2,7 @@
 name = TextMate
 
 [configuration_files]
-Library/Application Support/TextMate
+Library/Application Support/TextMate/Global.tmProperties
+Library/Application Support/TextMate/Managed
 Library/Preferences/com.macromates.textmate.plist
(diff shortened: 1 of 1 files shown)


samsonjs:
Ugh, it's something in Library/Application Support/TextMate/Managed/. It will take a while to track down what is broken here.

samsonjs:
I'm not sure what it is. I diffed the folders but didn't come up with anything. This pull request doesn't fix it though.

lra:
Feel free to ask the devs

kfinlay:
This behavior led me to switch from TextMate to Sublime Text, which works very well under syncing. I could never figure out a custom syncing solution that didn't eventually break TextMate.

PR #541, opened 2015-05-12
Added the new default path for google drive

jaredbonobos:


lra:
Fixed in https://github.com/lra/mackup/pull/575

Issue #512, opened 2015-04-14
Should not back up private SSH keys by default


lazerwalker:
The default behavior of `mackup` currently includes backing up the `~/.ssh` dir, including all SSH private keys. There are **serious** security implications to uploading your SSH private key unencrypted to a third-party server (e.g. Dropbox).

When I ran `mackup backup`, I had no idea that would happen, and certainly would not have expected it from a cursory reading of the README. This is completely unacceptable. I'm incredibly lucky that I did not have Dropbox.app running at the time, or else I would have felt the need to regenerate my private key, causing me a GREAT deal of hassle and making me INCREDIBLY angry at your software.

I'd **strongly** recommend that SSH backup be disabled by default. My ideal solution would probably involve mackup encrypting everything before moving/uploading (e.g. into a TrueCrypt volume), but given the current symlinking behavior I'd assume that would involve a fair bit of work. 

Some users may be okay with the current behavior. That's fine, but it's not a decision to take lightly, and one that many users will decidedly NOT want. Whatever the solution looks like, this absolutely needs to be an opt-in behavior (if it exists at all).

vitorgalvao:
This has been repeated [time](https://github.com/lra/mackup/issues/336) and [time](https://github.com/lra/mackup/issues/109) again. There should be at least one more issue about this. @lra has also stated that syncing his ssh keys is part of the reason he started mackup in the first place, so this is unlikely to change.

Why these issues are still open, I have no idea.

lra:
You are right, I'll try to keep the discussion centralized.
Regarding the matter, I'm sold to the idea that it should not sync it by default, we just need to name the enabling/disabling of apps simple enough so the empty default application list can be easily modified.

Let's follow up at https://github.com/lra/mackup/issues/109

aristidesfl:
Mackup should actually not move any files without confirming with the user first. It should present the list of apps or files which will be backed up and confirm if the user wants to proceed.

lra:
> Mackup should actually not move any files without confirming with the user first.

That's an opinion, check the issues and some people want the opposite: to not have to be bothered.

aristidesfl:
How often do you run mackup backup?

lra:
once per computer + on every new app installed.

enricodeleo:
I found Mackup back in the days during a research about ssh key syncing :)

[...]

Issue #109, opened 2013-08-16
Don't touch .ssh and .gnupg with the default settings …


a7p:
putting these files into an unencrypted Dropbox-Folder is irresponsible - mackup should not be irresponsible for me (and I should read documentation more careful - I know).

doctaphred:
:+1: Just discovered this behavior after the fact, and I feel pretty violated. Even looking just now, I couldn't find anywhere in the docs that mention that this program uploads private keys by default—I certainly don't consider them "pure configuration files". Glad to see you're sold on the idea of making it non-default (#512), but please at least note this behavior in the main README and `mackup -h` until it's implemented.

Other than that, this is an awesome utility!

MinimaxHeadroom:
I agree with all the comments above. Not clearly and *prominently* listing in the main documentation that such sensitive information as a user's .ssh folder would automatically be synced to the cloud—regardless of whether or not the user's private keys were password-protected—is incredibly careless.

Please omit this default behaviour or, at a minimum, include a prominent disclaimer.

sqren:
I've now had to void my private key twice because of the auto-sync. I really like Mackup but don't this is a sensible default.

lra:
TODO https://github.com/lra/mackup/issues/606

lra:
Added a TODO for encryption: https://github.com/lra/mackup/issues/606

lra:
just merged a PR to not sync ssh keys by default https://github.com/lra/mackup/pull/295
if anyone wants to do one for GnuPG, you are welcome.

PR #545, opened 2015-05-17
Add warnings to README and --help about syncing private keys.

doctaphred:
This should really become non-default behavior, per #109 and #512, but
in the meantime let's at least warn users who might not like it.


bjeavons:
:+1: :100: :exclamation:

PR #547, opened 2015-05-19
Python3

TrigonaMinima:
Python3 strings and input function. I have ```Python 3.4.3 ``` installed using anaconda on my system.
When I ran ```mackup backup```, it gave an error that the Dropbox directory didn't exist which was happening due to the byte string. This pull request takes care of the same.

### DIFF
--- a/mackup/config.py
+++ b/mackup/config.py
@@ -2,6 +2,7 @@
 
 import os
 import os.path
+import sys
 
 from .constants import (MACKUP_BACKUP_PATH,
                         MACKUP_CONFIG_FILE,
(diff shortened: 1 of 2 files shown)


dufferzafar:
:+1:

mackup/config.py
```diff
@@ -206,7 +207,11 @@ def _parse_path(self):
                 raise ConfigError("The required 'path' can't be found while"
                                   " the 'file_system' engine is used.")
 
-        return str(path)
+        # Python 2 and python 3 byte strings are different.
+        if sys.version_info[0] < 3:
+            return str(path)
+        else:
+            return path.decode("utf-8")
```
lra: This function is supposed to return a `str`, see docstring

TrigonaMinima: Yeah, even now a string is being returned. Previously this was not working. This might make things a bit more clearer of how byte strings are handled in python3.

```
>>> var = b'abc'
>>> var
b'abc'
>>> str(var)
"b'abc'"
>>> var.decode("utf-8")
'abc'
```

lra: ok thats fine.
Sorry to be a bother, but can you keep 1 return at the end of the function? Assign a temp var, thats fine.

TrigonaMinima: Sure.

PR #551, opened 2015-05-30
The current version of Mailplane has a "3" in the filename plist

roderik:


lra:
We still need to support the old version, can you add back the previous file?
Also mind adding a changelog entry?
Thanks

roderik:
Done, should have read the contributing guide :)

PR #555, opened 2015-06-02
Improved the current transmission support.

TrigonaMinima:
Now, working in Linux where configurations are saved in ```.config/transmission```.

### DIFF
--- a/mackup/applications/transmission.cfg
+++ b/mackup/applications/transmission.cfg
@@ -5,7 +5,5 @@ name = Transmission
 Library/Preferences/org.m0k.transmission.plist
 Library/Application Support/Transmission/blocklists
 .config/transmission/blocklists
-.config/transmission/resume
-.config/transmission/torrents
 .config/transmission/settings.json
 .config/transmission/stats.json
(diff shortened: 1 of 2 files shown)


mackup/applications/transmission.cfg
```diff
@@ -4,3 +4,8 @@ name = Transmission
 [configuration_files]
 Library/Preferences/org.m0k.transmission.plist
 Library/Application Support/Transmission/blocklists
+.config/transmission/blocklists
+.config/transmission/resume
+.config/transmission/torrents
```
lra: Can you `ls -l` those folders? You don't want to sync binary files, e.g. torrents.

TrigonaMinima: ```.config/transmission/blocklists``` is same as 
```Library/Application Support/Transmission/blocklists```.

There are text files in ```.config/transmission/resume``` but ```.config/transmission/torrents``` contains binary files. I just included both because in the new setup of ```transmission``` to resume I'll need to get the same torrent files. So, I included both folders.

lra: Sorry but mackup is not a NAS, it's for storing config, not data.
If you do this, your dropbox will constantly sync.

TrigonaMinima: Ohh! Got you. Sorry, didn't think about that.

PR #472, opened 2015-02-17
Update git.cfg

zanderzhng:
adding .gitignore_global file located in ~

PR #557, opened 2015-06-02
Improved git support for .gitignore_global

TrigonaMinima:


lra:
It's not an official file https://github.com/lra/mackup/pull/472

TrigonaMinima:
It still is a configuration file. At least, I won't have to explicitly copy it. Wasn't that the whole goal of Mackup?
Anyway, you can close the PR if you want. :)

lra:
If everyone comes with his own file to sync in the config, we'll end up with zillion files that make sense for no one.
The fine line is, if it's not official, it needs to go in the user config: https://github.com/lra/mackup/tree/master/doc#add-support-for-an-application-or-any-file-or-directory

PR #564, opened 2015-06-06
Added .zshrc to Oh My Zsh

ronniemoore:
Added .zshrc to Oh My Zsh
  (.zshrc is referenced several times in [docs](https://github.com/robbyrussell/oh-my-zsh#enabling-plugins))


jannae:
This file is already accounted for in the zsh.cfg file

https://github.com/lra/mackup/blob/a030e314be0b1e51ec842197c13a0fb976cbbccb/mackup/applications/zsh.cfg

ronniemoore:
Thanks!

PR #567, opened 2015-06-09
OSX Services Workflows

TCattd:
Example for what it's backing up
http://brettterpstra.com/howtos/install-an-os-x-system-service/


lra:
I feel like that's covered by user configuration, and should not live in the default distribution.

PR #569, opened 2015-06-09
Mailplane 3 improved support

TCattd:


lra:
Isn't it the folder that stores a lot of files?
Also merge conflicts.

TCattd:
That path keeps two xml, Accounts.xml and WorkspaceStates.xml. Accounts keeps the logged in account, and the other xml the current state of tabs inside mailplane; don't know if the last is needed or can cause some conflict.

And the password account is inside the keychain, not in those xml https://twitter.com/mailplane/status/324561904381079555

That path contains a localStorageDatabase/ folder inside too. It only has some .localstorage cache items http://i.imgur.com/tY7vNl2.png from mailplane's internal browser implementation.

Maybe just the Accounts.xml is needed, so you don't have to re-add your mail account after you install the app (as happened to me here at least). Can change the PR if is needed.

lra:
ok up to you, can you solve the conflicts and I'll merge

TCattd:
Done.

PR #570, opened 2015-06-17
Add .rspec to ruby.cfg

DomKM:


lra:
isn't it specific to rspec? http://rspec.info/

DomKM:
Yes, it is specific to RSpec. Mackup's Ruby config already supports `.pryrc` which is specific to [Pry](http://pryrepl.org/)

lra:
then it's a mistake, feel free to open a PR to fix those =)

PR #572, opened 2015-06-18
Add support for Glyphs 2.0 - www.glyphsapp.com

drawcard:


mackup/applications/glyphs2.cfg
```diff
@@ -0,0 +1,7 @@
++[application]
```
lra: ++?

lra:
https://github.com/lra/mackup/blob/master/CONTRIBUTING.md

PR #596, opened 2015-08-04
support for iCloud drive as an engine

jlowin:
If it's called Mackup, it has to support iCloud Drive out of the box, right?

On Yosemite, iCloud Drive is available as a folder with a slightly convoluted path. This exposes it as an engine. The alternative is to use a filesystem engine with that same path.


lra:
that makes sense, thanks! But WTF is that ugly path?! Apple...
I'll test it, unless someone else wants to =)

mkwmms:
Tested it on OS X 10.11 beta 5 and OS X 10.10.4 and it works great. Awesome PR!

mackup/utils.py
```diff
@@ -267,6 +267,23 @@ def get_copy_folder_location():
...
 
 
+def get_icloud_folder_location():
+    """
+    Try to locate the iCloud Drive folder.
+
+    Returns:
+        (str) Full path to the iCloud Drive folder.
+    """
+    yosemite_icloud_path = '~/Library/Mobile Documents/com~apple~CloudDocs/'
```
lra: I don't trust python to be able to expand `~` in all envs, can you use the HOME env vars like in other get_*_location to be sure it works everywhere?

lra: Oh I see, u use expanduser, that's fine

lra:
nice work!

PR #597, opened 2015-08-04
Added Library/Application Support/LightTable/User (user.keymap, etc)

Joshfindit:
Most importantly, this folder holds
    user.behaviors
    user.keymap

### DIFF
--- a/mackup/applications/light-table.cfg
+++ b/mackup/applications/light-table.cfg
@@ -4,4 +4,5 @@ name = Light Table
 [configuration_files]
 Library/Application Support/LightTable/plugins
 Library/Application Support/LightTable/settings
+Library/Application Support/LightTable/User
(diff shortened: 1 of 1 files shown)


lra:
anything else big in this folder?

Joshfindit:
Here are the contents:

```
plugin.edn
src/lt/plugins/user.cljs
user.keymap
project.clj
user.behaviors
user_compiled.js
```

I'm not personally sure what all these files do, but according to documentation the settings for keymaps, workspaces, and behaviours are all stored here.
After doing a cursory test with plugin.edn, it updates when plugins are installed by the user.
I would also suspect (but cannot confirm) that some users would create new files in Library/Application Support/LightTable/User to extend LightTable.

lra:
ok cool, as long as there is not something insane in it.

PR #599, opened 2015-08-08
Added support for httpie

amatos:


lra:
Shouldn't it be just the config files?
https://github.com/jkbrzt/httpie#config
Looks like this folder can have session stuff too.

PR #600, opened 2015-08-08
make parent dirs when creating mackup home

mkwmms:
This allows you to specify something like:
[storage]
engine = icloud
directory = .config/mackup


lra:
cool

PR #616, opened 2015-08-10
Added .gitconfig_global configuration

rstuven:


lra:
Can you point to an official mention of this file in the git doc or source?
If not, it's a user config.

Issue #618, opened 2015-08-10
Reference shell environment variables from within app.cfg files


mkwmms:
Is it possible/would it be possible to reference `$XDG_CONFIG_HOME` etc. environment variables from within app.cfg files?

For example:
My `$XDG_CONFIG_HOME` is set to my iCloud Drive directory/.config (aka the ugly path: `~/Library/Mobile Documents/com~apple~CloudDocs/.config`)

So for .pythonrc it looks like:
```
[application]
name = Python

[configuration_files]
Library/Mobile Documents/com~apple~CloudDocs/.config/python/.pythonrc
```
But it would be awesome if it could look like:
```
[application]
name = Python

[configuration_files]
$XDG_CONFIG_HOME/.config/python/.pythonrc
```
or:
`os.environ['XDG_CONFIG_HOME']/python/.pythonrc`

This would also be helpful for apps that support custom environment variables like atom:
```
[application]
name = Atom

[configuration_files]

Library/Preferences/com.github.atom.plist
$ATOM_HOME/config.cson
$ATOM_HOME/init.coffee
$ATOM_HOME/keymap.cson
$ATOM_HOME/keymaps
$ATOM_HOME/snippets.cson
$ATOM_HOME/styles.less
```
__default app XDG support__
This would also be helpful for apps that support the XDG spec like git (see: the [git spec](http://git-scm.com/docs/git-config))
```
[application]
name = Git

[configuration_files]
$XDG_CONFIG_HOME/git/config
$XDG_CONFIG_HOME/git/credentials
$XDG_CONFIG_HOME/git/ignore
.gitignore
.gitcredentials
.gitconfig
```

lra:
In your case it would work, but what about people that don't define `$XDG_CONFIG_HOME` ?

mkwmms:
Am I wrong in thinking that you can specify multiple directory/files that "overlap" and mackup will only handle the ones that exist? 

Or are you saying mackup will throw an exception etc. if the variable is undefined rather than expanding it first, if set, and then checking the directories...

```
[application]
name = Git

[configuration_files]
$XDG_CONFIG_HOME/git/config
$XDG_CONFIG_HOME/git/credentials
$XDG_CONFIG_HOME/git/ignore
.gitignore
.credentials
.gitconfig
``` 
because it would be useless to specify both `$XDG_CONFIG_HOME/git/config` and a `.gitconfig` so mackup would backup either/or, but not both

mkwmms:
[...]

PR #624, opened 2015-08-15
XDG support for app configs

mkwmms:
adds XDG support for app config files as discussed in #618

### DIFF
--- a/mackup/appsdb.py
+++ b/mackup/appsdb.py
@@ -73,7 +73,8 @@ class ApplicationsDatabase(object):
                                     .format(path))
                             path = os.path.join(xdg_config_home, path)
                             path = path.replace(home, '')
-                            self.apps[app_name]['configuration_files'].add(path)
+                            self.apps[app_name]['configuration_files'] \
+                                    .add(path)
 
     @staticmethod
     def get_config_files():
--- a/mackup/appsdb.py
+++ b/mackup/appsdb.py
@@ -45,13 +45,29 @@ class ApplicationsDatabase(object):
                 self.apps[app_name]['name'] = app_pretty_name
 
                 # Add the configuration files to sync
-                self.apps[app_name]['configuration_files'] = set()
-                if config.has_section('configuration_files'):
-                    for path in config.options('configuration_files'):
+                config_section = 'configuration_files'
+                if config.has_section(config_section):
+                    self.apps[app_name][config_section] = set()
+                    for path in config.options(config_section):
                         if path.startswith('/'):
                             raise ValueError('Unsupported absolute path: {}'
                                              .format(path))
-                        self.apps[app_name]['configuration_files'].add(path)
+                        self.apps[app_name][config_section].add(path)
+
+                # Add the XDG configuration files to sync
+                xdg_config_section = 'xdg_configuration_files'
+                xdg_config_home = os.environ.get('XDG_CONFIG_HOME', False)
+                if config.has_section(xdg_config_section):
+                    self.apps[app_name][xdg_config_section] = set()
+                    if xdg_config_home:
+                        for path in config.options(xdg_config_section):
+                            if path.startswith('/'):
+                                raise ValueError(
+                                    'Unsupported absolute path: {}'
+                                    .format(path))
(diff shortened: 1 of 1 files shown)


mackup/appsdb.py
```diff
@@ -45,13 +45,29 @@ def __init__(self):
                 self.apps[app_name]['name'] = app_pretty_name
 
                 # Add the configuration files to sync
-                self.apps[app_name]['configuration_files'] = set()
-                if config.has_section('configuration_files'):
-                    for path in config.options('configuration_files'):
+                config_section = 'configuration_files'
+                if config.has_section(config_section):
+                    self.apps[app_name][config_section] = set()
```
lra: the set is to be defined even if the section does not exist.

mackup/appsdb.py
```diff
@@ -45,13 +45,29 @@ def __init__(self):
...
+                    for path in config.options(config_section):
                         if path.startswith('/'):
                             raise ValueError('Unsupported absolute path: {}'
                                              .format(path))
-                        self.apps[app_name]['configuration_files'].add(path)
+                        self.apps[app_name][config_section].add(path)
+
+                # Add the XDG configuration files to sync
+                xdg_config_section = 'xdg_configuration_files'
+                xdg_config_home = os.environ.get('XDG_CONFIG_HOME', False)
```
lra: you need to make sure XDG_CONFIG_HOME is under the user's home as stated in https://github.com/lra/mackup/issues/618#issuecomment-129642503 and fail fast.

lra: Just do `os.environ.get('XDG_CONFIG_HOME')` it's returning `None` if not set.

mackup/appsdb.py
```diff
@@ -45,13 +45,29 @@ def __init__(self):
...
+                if config.has_section(xdg_config_section):
```
lra: check that there is a `xdg_config_home` 1st.

mackup/appsdb.py
```diff
@@ -45,13 +45,29 @@ def __init__(self):
...
+                    self.apps[app_name][xdg_config_section] = set()
```
lra: why a new set? if there are xdg paths to treat, they must be added to the `configuration_files` set

mackup/appsdb.py
```diff
@@ -45,13 +45,29 @@ def __init__(self):
...
+                    if xdg_config_home:
+                        for path in config.options(xdg_config_section):
+                            if path.startswith('/'):
+                                raise ValueError(
+                                    'Unsupported absolute path: {}'
+                                    .format(path))
+                            path = os.path.join(xdg_config_home, path)
+                            path = path.replace(os.path.expanduser('~/'), '')
```
lra: no, we don't support this.

mackup/appsdb.py
mkwmms: os.environ.get('XDG_CONFIG_HOME', False) 
False is returned if `$XDG_CONFIG_HOME` is not set, thus it won't enter the if statement on line 61 (which is failing fast of sorts...)

mackup/appsdb.py
mkwmms: Ok, I had them being added to the same set before and then changed it because I wasn't sure if you would want them separate in the AppsDB... Ill change it back :)

mkwmms:
I kept running into it not passing formatting checks because the lines were 80 characters (exceeding the max of 79) and that is why I was doing weird things with `self.apps[app_name][config_section].add(path)` and `self.apps[app_name]['configuration_files'].add(path)` and the same with XDG to make the lines shorter but consistent between XDG and regular configuration_files...

lra:
tests are failing

mkwmms:
> tests are failing

Sorry, I didn't know how you felt about explicit line continuation. I fixed it (and your other suggestions, to my knowledge, though.

mkwmms:
Thanks for the quick merge! 

Love this app!

lra:
Thanks for the PR, feel free to update the doc to explain this new feature and the way to enable this.

mkwmms:
Will do.

PR #629, opened 2015-08-22
git.cfg XDG support 

mkwmms:
I would love to also add:
```
[xdg_configuration_files]
git/ignore
```
I know there was some talk about `git/ignore` being overwritten but I have never had that issue when using it from `$XDG_CONFIG_HOME` rather than `~/.config`

### DIFF
--- a/mackup/applications/git.cfg
+++ b/mackup/applications/git.cfg
@@ -3,3 +3,7 @@ name = Git
 
 [configuration_files]
 .gitconfig
+
+[xdg_configuration_files]
+git/config
+git/credentials
--- a/mackup/applications/git.cfg
+++ b/mackup/applications/git.cfg
@@ -6,4 +6,3 @@ name = Git
 
 [xdg_configuration_files]
 git/config
-git/credentials


mackup/applications/git.cfg
```diff
@@ -3,3 +3,7 @@ name = Git
 
 [configuration_files]
 .gitconfig
+
+[xdg_configuration_files]
+git/config
+git/credentials
```
lra: not sure people will like this, I've got a lot of bad feedback from trying to sync ssh keys by default.

mkwmms: Fair enough, although they aren't keys...

The entirety of mine is:
```
[credential]
	helper = osxkeychain

[user]
	name = William Myers
        email=
```

It is usually used so people can share .gitconfig files and keep their emails etc. separate...

But I completely understand if you just want to do:

```
[xdg_configuration_files]
git/ignore
```

lra: > Using this helper will store your passwords unencrypted on disk, protected only by filesystem permissions.

http://git-scm.com/docs/git-credential-store

mkwmms: I guess I have just been using it to tell Git to store them in OS X's Keychain.
Ill take it out as I agree with you on not syncing credentials.

PR #650, opened 2015-09-19
Add Sketch

rstacruz:
I hope I did this right. I'm not sure where Sketch stores templates though. This should also ignore ~/Library/Application Support/com.bohemiancoding.sketch3/.license with should be unique per machine.

http://bohemiancoding.com/

### DIFF
--- a/mackup/applications/sketch
+++ b/mackup/applications/sketch
@@ -1,0 +1,6 @@
+[application]
+name = Sketch
+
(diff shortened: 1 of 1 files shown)


vitorgalvao:
https://github.com/lra/mackup/pull/212.

rstacruz:
oh, that's even better

PR #664, opened 2015-10-04
add support for 1password backups

zanderzhng:
this syncs 1password's backups so people won't forget to backup these backups.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Add support for Blackfire (via @mnapoli)
 - Add support for liquidprompt (via @GochoMugo)
 - Add support for REdshift (via @orschiro)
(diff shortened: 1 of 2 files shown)


lra:
If we say that mackup officially supports 1password: 2 things that are not straightforward and might confuse people:
1. On my App Store 1password, the backups are in the containers of 1password, which is a dynamic path, e.g. `/Users/analogue/Library/Containers/2BUA8C4S2C.com.agilebits.onepassword-osx-helper/Data/Library/Backups`
1. 1password already supports dropbox and all, there might be a reason why that don't put the backups in the synced filesystem (e.g. dropbox)

For those reasons, I think it might be more relevant as a user config, see https://github.com/lra/mackup/tree/master/doc#add-support-for-an-application-or-any-file-or-directory but if more people think it's a good idea, why not.

Issue #615, opened 2015-08-09
Support for Mac App Store applications


lra:
From @vitorgalvao "For MAS apps, it’ll likely also need to defaults read preferences to a file and restore them with defaults write. This is why I’ve been mentioning storing bundle IDs."

Depends on https://github.com/lra/mackup/issues/613 1st.

PR welcome

lra:
Not that I know of, sadly.

PR #670, opened 2015-10-11
Add support for Pocket

rbartoli:


lra:
Did you try it? It should not work.
App Store apps are not supported yet.
See https://github.com/lra/mackup/issues/615

PR #673, opened 2015-10-11
Add support for Electrum

rbartoli:


lra:
Not sure people will appreciate having their bitcoin wallet on dropbox.
Is it intended? (e.g. encrypted)

rbartoli:
Hi @lra,

It seems that the default_wallet is encrypted by default, however I think is worth contacting the electrum developer to confirm that is a good idea to backup both files. I'll update the PR based on his feedback.

rbartoli:
Following my conversation with @dabura666, it seems that the `default_wallet` is not encrypted by default unless the user sets a password (this is not required).

So it's a good idea to leave the `default_wallet` out. I've updated the PR.

Issue #339, opened 2014-10-04
Chinese Character in the Path of Google Drive Home


digglife:
The default path of Google Drive Home on my system is `~/Google 云端硬盘`, which means `Google cloud disk` in Chinese. When I fire up `mackup backup`, the following error occurs. 

``` python
Traceback (most recent call last):
  File "/usr/local/bin/mackup", line 26, in <module>
    mackup.main.main()
  File "/usr/local/Cellar/mackup/0.7.3/lib/python2.7/site-packages/mackup/main.py", line 38, in main
    mckp = Mackup()
  File "/usr/local/Cellar/mackup/0.7.3/lib/python2.7/site-packages/mackup/mackup.py", line 21, in __init__
    self._config = config.Config()
  File "/usr/local/Cellar/mackup/0.7.3/lib/python2.7/site-packages/mackup/config.py", line 43, in __init__
    self._path = self._parse_path()
  File "/usr/local/Cellar/mackup/0.7.3/lib/python2.7/site-packages/mackup/config.py", line 188, in _parse_path
    return str(path)
UnicodeEncodeError: 'ascii' codec can't encode characters in position 21-24: ordinal not in range(128)
```

Apparently it's because the path of Google Drive Home is [unicode](https://github.com/lra/mackup/blob/master/mackup/utils.py#L245), but here `str()` tries to encode it with ASCII.

I got a quick fix but it's urgly :(. Change the return value of `get_google_drive_folder_location()` to byte string. 

``` python
-googledrive_home = unicode(data[0])
+googledrive_home = data[0].encode('utf-8')
```

Now it's fine for mackup to recognize the path.

``` bash
zhumac:Downloads erichu$ mackup backup
Mackup needs a directory to store your configuration files
Do you want to create it now ? </Users/erichu/Google 云端硬盘/Mackup> <Yes|No>
```

PR #683, opened 2015-10-19
Fix for non-ascii character in the path of backup engine (issue #339)

digglife:
#339

The basic idea is using bytes string under py2, using string(unicode) under py3. Although this fix is for Google Drive at the first, but it works for other engines as well. I also modified the test files a bit for testing the changes under both py2 and py3. For more details, please refer to the inline comments.


mackup/utils.py
```diff
@@ -238,7 +241,7 @@ def get_google_drive_folder_location():
                      "WHERE entry_key = 'local_sync_root_path';")
             cur.execute(query)
             data = cur.fetchone()
-            googledrive_home = unicode(data[0])
+            googledrive_home = data[0]
```
digglife: `unicode` is not a valid method under py3. It returns unicode string anyway.

mackup/config.py
```diff
@@ -218,11 +218,10 @@ def _parse_path(self):
                 raise ConfigError("The required 'path' can't be found while"
                                   " the 'file_system' engine is used.")
 
-        # Python 2 and python 3 byte strings are different.
+        # For py2, encode the path with utf8 to bytes string because
+        # there might be non-ascii characters in it.
         if sys.version_info[0] < 3:
-            path = str(path)
-        else:
-            path = path.decode("utf-8")
```
digglife: under py3, `decode` method is only valid when path is `bytes` type, but path can be either `str` or `bytes` in the original code.

tests/config_tests.py
```diff
@@ -9,6 +10,11 @@
                               ENGINE_FS)
 from mackup.config import Config, ConfigError
 
+def is_equal_string(src_path, dst_path):
+    if sys.version_info[0] < 3:
+        src_path = src_path.decode('utf8')
+    return src_path == dst_path
+
```
digglife: `decode` the bytes string path for comparing with unicode string in tests(py2).

mackup/utils.py
```diff
@@ -206,7 +209,7 @@ def get_dropbox_folder_location():
             data = f_hostdb.read().split()
     except IOError:
         error("Unable to find your Dropbox install =(")
-    dropbox_home = base64.b64decode(data[1])
+    dropbox_home = base64.b64decode(data[1]).decode('utf8')
```
digglife: `base64.b64decode(data[1])` returns bytes under py3. `decode` here for getting a unicode string, no matter the environmet is  py2 or py3.

rposborne:
This is causing some issues as it will prevent someone from uninstalling mackup, could we get this rebased on master and merged?

PR #686, opened 2015-10-19
Add support for GIMP

mateusrevoredo:


lra:
I'm scared that the synced folder is gonna have a lot of cache/binary stuff. Can you dig in?

mateusrevoredo:
Yes!
In case the folder contains lots of unnecessary stuff, can I narrow the configuration file to only sync the absolutely necessary files?

mateusrevoredo:
As you can see, the synced files are shortcuts, plugins, scripts, templates themes etc.All of those files are part of someone's configuration.
The only exception would be the temp folder. But it is only used when exporting an image. It creates a file and then deletes the file after the export operation ends.
In my Mac the folder currently “weighs” ~700 Kb.

gimp01
gimp02

lra:
Try it locally and let us know.

mateusrevoredo:
I've been using this for more than a week and it's working like expected.

fharper:
@mateusrevoredo: any plans to update this PR and merge it?

PR #700, opened 2015-10-31
Alpha sort backup, restore, uninstall

Cottser:
Starting to play around with mackup again and was testing the verbose and dry run options. I noticed that the output when running the main commands is in a somewhat arbitrary order (didn't track down exactly why, didn't seem to be because os.listdir is arbitrary but maybe just because it's a set), I think mackup's main commands should be sorted like `list` is.

### DIFF
--- a/mackup/main.py
+++ b/mackup/main.py
@@ -81,7 +81,7 @@ def main():
         mckp.check_for_usable_backup_env()
 
         # Backup each application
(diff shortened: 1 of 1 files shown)


lra:
Sorry, merge conflicts, I just had a bug to fix.

Cottser:
No worries I thought that looked a bit off when I was looking at it, glad it got squashed. Updated.

PR #701, opened 2015-11-01
Windows support

neutralevil:
It is expected to work properly on Windows Vista or Higher with Python version 3.2+.
But many testcases can't pass, cos they heavily rely on Python 2 and Unix.


neutralevil:
Oops, coverage failing, but how can I write testcase running on Linux to cover Windows specific code?

lra:
I guess you'll have to mock.
Also I suggest doing 1 PR for python 3.2, and one for Windows

PR #703, opened 2015-11-01
Fix for neovim config path change.

jiegec:
Neovim now changes to XDG configuration path.


lra:
Can you point to some official doc or source code stating that the old path are not supported anymore?

jiegec:
@lra https://github.com/neovim/neovim/wiki/Following-HEAD

PR #733, opened 2015-12-14
Leiningen support

thirteen37:
Support for Clojure's defacto build tool, [Leiningen](http://leiningen.org/)


mackup/applications/leiningen.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = Leiningen
+
+[configuration_files]
+.lein
```
lra: Can you end the last line with a newline character?

PR #739, opened 2015-12-19
Add Tunnelblick support

sidick:
I'm unsure if you'll want to merge this as it does sync VPN certificates, but really they should be password protected anyway, so it's up to you :)


ivan-leschinsky:
I like it!

PR #743, opened 2015-12-31
added support for fasd

doubleloop:
added support for [fasd](https://github.com/clvv/fasd)

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,7 @@ See the [README](doc/README.md) file in the doc directory for more info.
 - [ExpanDrive](http://www.expandrive.com/)
 - [Fantastical](http://flexibits.com/fantastical)
 - [Feeds](http://www.feedsapp.com/)
+- [fasd](https://github.com/clvv/fasd)
 - [Filezilla](https://filezilla-project.org/)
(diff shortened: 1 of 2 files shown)


SharkIng:
Is this bugged? everytime I add new configure, i am running mackup backup to add new configure into mackup backup. However, all other thing are OK but .fasd as me everytime when i run the command. seem like it try to back it up everytime. Why?

doubleloop:
True, I verified it just now.. it seams that fasd is overwriting the file instead of appending to it.

https://github.com/clvv/fasd/blob/master/fasd#L368

I will try to make PR to fasd to fix it.

doubleloop:
@SharkIng 
Since my PR to fasd is still not merged, I can share a simple work around, you can just create a `.fasdrc` file with:
```
FASD_DATA="$HOME/.fasd/fasd"
```
then if you move the .fasd data files to a directory, fore example like this:
```
mkdir ~/tmpdir
mv ~/.fasd ~/tmpdir/fasd
mv ~/tmpdir ~/.fasd
```
the whole folder will be symlinked.

PR #744, opened 2015-12-31
added linux support for Audacious

doubleloop:
added linux support for [Audacious](http://audacious-media-player.org/)

### DIFF
--- a/mackup/applications/audacious.cfg
+++ b/mackup/applications/audacious.cfg
@@ -1,0 +1,9 @@
+[application]
+name = Audacious
+
+[configuration_files]
+.config/audacious/config
+.config/audacious/plugin-registry
+.config/audacious/playlists
+.config/audacious/playlist-state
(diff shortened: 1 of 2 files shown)


mackup/applications/audacious.cfg
```diff
@@ -0,0 +1,9 @@
+[application]
+name = Audacious
+
+[configuration_files]
+.config/audacious/config
+.config/audacious/plugin-registry
+.config/audacious/playlists
+.config/audacious/playlist-state
+.config/audacious/scrobbler.log
```
lra: Not sure we want to sync a log file, do we?

doubleloop:
It is for slastfm (http://www.last.fm) scrobbler plugin so that if we are offline and listen to musicte scrobbles are stored in this file to not be lost and in case of getting online to be read from

lra:
I see, thanks

PR #754, opened 2016-01-07
backup user level gitignore

CatTail:


lra:
Until this is an official file, this should go in you user config, see https://github.com/lra/mackup/pull/498

Issue #758, opened 2016-01-18
Byobu - trying to backup a socket for SSH agent gives error message


neil-greenwood:
If your Byobu configuration contains a socket connected to the local SSH agent, you get an error message when running `mackup backup`:

```
Backing up .byobu ...
Traceback (most recent call last):
  File "/usr/local/Cellar/mackup/0.8.12/libexec/bin/mackup", line 9, in <module>
    load_entry_point('mackup==0.8.12', 'console_scripts', 'mackup')()
  File "/usr/local/Cellar/mackup/0.8.12/libexec/lib/python2.7/site-packages/mackup/main.py", line 90, in main
    app.backup()
  File "/usr/local/Cellar/mackup/0.8.12/libexec/lib/python2.7/site-packages/mackup/application.py", line 113, in backup
    utils.copy(home_filepath, mackup_filepath)
  File "/usr/local/Cellar/mackup/0.8.12/libexec/lib/python2.7/site-packages/mackup/utils.py", line 105, in copy
    shutil.copytree(src, dst)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/shutil.py", line 208, in copytree
    raise Error, errors
shutil.Error: [('/Users/neil/.byobu/.ssh-agent', '/Users/neil/./Mackup/.byobu/.ssh-agent', "[Errno 102] Operation not supported on socket: '/Users/neil/.byobu/.ssh-agent'")]
```

lra:
should be fixed by https://github.com/lra/mackup/pull/759

Issue #479, opened 2015-03-06
No graceful error management


sieben:
I get the following error out of the box:

shutil.Error: [('/home/sieben/.byobu/.ssh-agent', '/home/sieben/Dropbox/Mackup/.byobu/.ssh-agent', "[Errno 2] No such file or directory: '/home/sieben/.byobu/.ssh-agent'")]

lra:
What kind of graceful error management would you expect to see?

neil-greenwood:
I think this file should be excluded. I have a patch to ignore that file. I'm just preparing the PR now...

lra:
should be fixed by https://github.com/lra/mackup/pull/759

PR #759, opened 2016-01-18
Specify Byobu files except SSH agent socket.

neil-greenwood:
This fixes issue #758 and is a work-around for issue #479.


lra:
thanks!

PR #761, opened 2016-01-19
Add support for backing up keyboard layouts

oillio:


lra:
OSX stuff usually does not like being linked. Can anyone else confirm this is working?

PR #762, opened 2016-01-20
Add support for LightPaper

robmoggach:
[It](http://lightpaper.42squares.in/) is lovely. [LightPaper](http://lightpaper.42squares.in/), that is. Now it's all mack'd up.

### DIFF
--- a/mackup/applications/lightpaper.cfg
+++ b/mackup/applications/lightpaper.cfg
@@ -1,0 +1,7 @@
+[application]
+name = LightPaper
+
+[configuration_files]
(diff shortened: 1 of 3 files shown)


mackup/applications/lightpaper.cfg
```diff
@@ -0,0 +1,7 @@
+[application]
+name = LightPaper
+
+[configuration_files]
+Library/Preferences/com.42squares.LightPaper.plist 
+.lightpaper
+Library/Application\ Support/LightPaper/customCSS.css
```
lra: check the other cfg files, no need for the `\`

PR #765, opened 2016-01-31
Add support for TextMate 2

wchrisjohnson:
Add support for TextMate 2


lra:
Do you mind merging this into `textmate.cfg`?

wchrisjohnson:
I don't think this is necessary as Textmate 2 appears to be supported implicitly.

PR #766, opened 2016-02-04
Add .vim/plugin/settings folder

gmpetrov:


mackup/applications/vim.cfg
```diff
@@ -14,6 +14,7 @@ name = Vim
 .vim/indent
 .vim/syntax
 .vim/vimrc
+.vim/plugin/settings
```
lra: let's keep it a-z ordered to prevent duplicates.

Issue #705, opened 2015-11-04
Skype GIFs backed up


altosaar:
In my 'recent' dropbox items are hundreds of gifs mackup has backed up from skype. Would be nice to disable having these backed up :)

lra:
You need to improve https://github.com/lra/mackup/blob/master/mackup/applications/skype.cfg

anantshri:
Hey Folks, I am new to this is there a simpler guide on how to write cfg's also what i am getting is the following folder needs to be excluded.

inside the backup folder : Library/Application Support/Skype/<username>/media_messaging/emo_cache/ <- this folder needs to be excluded from the backup. But considering we are syncing the whole folder via symlink this sounds like a bit of work.

All i can think of is backing up all folders except this one but since it has a username element to it, this might not be as simple as just quoting it out

moltar:
Yes, super annouying. Clogs up the whole Google Drive sync with a bunch of useless files.

All of them named something like:

`^003AFEAFC0BBF745958FE4C77B695B0A493D4A48498992AF55^pdefault_5b000c95-f530-4a7f-b302-842675aaa249_distr`

All are useless GIFs, MP4s and PNGs. I don't even know why Skype comes bundled with this crap.

moltar:
The full location is:

`Library/Application Support/Skype/%SKYPE_USERNAME%/media_messaging/emo_cache`

moltar:
For now I simply deleted everything inside that folder and then changed permissions to make it non-writable. So far I can still see normal emoticons (smiles), and Skype has not complained.

lra:
can someone come up with a PR? I don't use Skype.

moltar:
I could, I guess. I'm just not sure of the syntax. Is there a way to blacklist a certain dir? Or do we just need to explicitly list every dir that is good?

lra:
> Is there a way to blacklist a certain dir?

No

PR #768, opened 2016-02-05
Skype fix issue 705

TCattd:
Fixes https://github.com/lra/mackup/issues/705

Since we can't blacklist a folder, and can't guess a user's folder (based on Skype user name), we backup just the content outside the user folder to avoid hundreds media files (like emoticons) pushed to Dropbox.
Not the best solution, but to avoid high CPU usage when using Skype, this is needed.

Tested.

My CPU usage wen't crazy because of this, in the middle of a meeting. Dropbox and Skype went berserker for those files. Could not use my laptop until i killed Skype and left Dropbox stop synchronizing all the assorted media Skype stores in there (username named folder).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - Add support for Rime (via @codefalling)
 - Improved support for R (via @wyf88)
 - Add support for Rhythmbox (via @orschiro)
+- Fixed Skype Support (via @TCattd)
 
 ## Mackup 0.8.12
 
--- a/mackup/applications/skype.cfg
+++ b/mackup/applications/skype.cfg
@@ -2,4 +2,6 @@
 name = Skype
 
 [configuration_files]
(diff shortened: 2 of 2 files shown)


lra:
This is great thanks.
Not sure if the folders you still sync are worth syncing, since I gave up on syncing Skype myself ;)
I'll let you decide as you use it.

TCattd:
Thanks.

Ideally, Skype shouldn't be synced at all, really.
I didn't know if a PR deleting skype.cfg entirely it had been understood :) that's why i just removed as i could the problematic syncing.

But if you want, @Ira, y can do another PR just deleting that file. No problem. There's really no point on sync anything on Skype just now.

lra:
I'd rather have someone take the decision, as I don't use it.
If anyone else want to jump in and make the call (PR), please do ;)

PR #778, opened 2016-03-09
Added support for Surge

soulboyx:
added support for Surge

### DIFF
--- a/mackup/applications/surge.cfg
+++ b/mackup/applications/surge.cfg
@@ -1,0 +1,5 @@
+[application]
+name = Surge
+
+[configuration_files]
+.surge.conf


lra:
Please, can you follow https://github.com/lra/mackup/blob/master/CONTRIBUTING.md ?

soulboyx:
Sorry that I missed the page. I did not know you would only like people to contribute following your todos.

I followed the link under CAN YOU SUPPORT APPLICATION X section to https://github.com/lra/mackup/tree/master/doc#get-official-support-for-an-application in the README for the project.

You might wanna update the file or the link so there will not be people doing the same thing again.

soulboyx:
I have to assume the Surge conf will never be supported in mackup since Surge only has a small group of users and it doesn't seem like this group would expand in the future. Since you would only like people to contribute by following your todos and tofixs, I doubt surge will be added to the todo list.

PR #782, opened 2016-03-19
Add support for IntelliJ IDEA 2016.1

jcgay:


mackup/applications/intellijidea-2016-1.cfg
```diff
@@ -0,0 +1,6 @@
+[application]
+name = IntelliJ IDEA 2016.1
+
+[configuration_files]
+Library/Application Support/IntelliJIdea2016.1
+Library/Preferences/IntelliJIdea2016.1
```
windwiny: The Community version name not include `IntelliJ`

    Library/Application Support/IdeaIC2016.1
    Library/Preferences/IdeaIC2016.1

jcgay: Ah thanks for the catch, I'm using the Ultimate edition.

jcgay:
Do you prefer to have two separate configurations (ultimate and community edition) or just merge everything into a single one ?

lra:
> Do you prefer to have two separate configurations (ultimate and community edition) or just merge everything into a single one ?

I think that would be great, care to do it?

jcgay:
Sure, done!

PR #787, opened 2016-03-22
moved intellij config to one file

doubleloop:


lra:
Dupe of https://github.com/lra/mackup/pull/782

PR #790, opened 2016-03-28
Added MPS Youtube support

fmartingr:
- Sync app configuration file
- Sync app local playlists file


fmartingr:
Solved conflicts.

mackup/applications/mps-youtube.cfg
```diff
@@ -0,0 +1,6 @@
+[application]
+name = MPS Youtube
+
+[configuration_files]
+.config/mps-youtube/config
+.config/mps-youtube/playlist_v2
```
vitorgalvao: It has a “no newline at end of file” warning. Those are [actually meaningful](http://unix.stackexchange.com/a/18789).

fmartingr: I can agree with you but I've seen some other config files without it so I'm leaving it as it is to match the current config files.

vitorgalvao: > but I've seen some other config files without it

Those are also wrong.

> I'm leaving it as it is to match the current config files.

That makes no sense. Some config files have this error, others do not, and you’re choosing to follow the objectively wrong behaviour.

The files that have this error should all be fixed.

fmartingr: It's not a big deal IMO, but I want to know what @lra have to say about it, because I prefer to do a separate PR adding all the obviously needed line breaks rather than just *doing it right* for this file only.

lra: > I've seen some other config files without it so I'm leaving it as it is to match the current config files.

Not sure what you refer to. All the cfg files end with a newline in this repo. @vitorgalvao's link explains why it's important.

> but I want to know what @lra have to say about it

Just thanks @vitorgalvao for catching this.

fmartingr: I'm sorry for my behaviour yesterday, I had a really bad day and acted like an idiot.
I've fixed the issue now, thanks @vitorgalvao for pointing it out.

lra:
thx

mackup/applications/mps-youtube.cfg
vitorgalvao: > I'm sorry for my behaviour yesterday, I had a really bad day and acted like an idiot.

No worries here, it’s all good. We disagreed, but it’s not like I thought you were being rude, or anything. The conversation remained civil, and that’s what matters.

PR #793, opened 2016-04-07
 Added support for Drush

penance316:


mackup/applications/drush.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = Drush
+
+[configuration_files]
+.drush
```
lra: can you finish this line with a newline char?

penance316: sure all done

PR #797, opened 2016-04-14
Adding support for Poet

leek:
Simply adding support for @awendt's [Poet](https://github.com/awendt/poet) tool.

### DIFF
--- a/mackup/applications/poet.cfg
+++ b/mackup/applications/poet.cfg
@@ -1,0 +1,5 @@
+[application]
+name = Poet
+
+[configuration_files]
+.ssh/config.d
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Added support for Drush (@penance316)
 - Added support for MPS-Youtube (via @fmartingr)
 - Add support for HandBrake (via @ryanjbonnell)
+- Add support for Poet (via @leek)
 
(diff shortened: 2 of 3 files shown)


mackup/applications/poet.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = Poet
+
+[configuration_files]
+.ssh/config.d
```
vitorgalvao: Wrong file.

leek: **@vitorgalvao:** Huh?

vitorgalvao: @leek Never mind. My mistake.

lra:
huh, sorry but this will lead to a lot of confusion.
`.ssh` is supposed to be owned by `ssh`, and this looks like a script hacking around its config files.
The problem is that if someone syncs `.ssh` it will mess things up.
This folder `.ssh/config.d` and everything in it should really be owned by `ssh`.

lra:
not sure how to solve this.
Move `.ssh/config.d` in `ssh`?

leek:
@lra I don't see any harm in having `.ssh/config.d` moved into `.ssh`.

leek:
I'm fine with this being in a custom `.cfg` for now. Closing this PR.

PR #802, opened 2016-04-19
Add .docker authentication config

gdraynz:


mackup/applications/docker.cfg
```diff
@@ -3,3 +3,4 @@ name = Docker
 
 [configuration_files]
 .dockercfg
+.docker
```
darcy: Seems like syncing the entire `.docker` folder is excessive as this includes all the docker machines as well, which can each be massive (over 30GB for me)

gdraynz: Damn, sorry about that, PR on the way

darcy: Nice! Thanks for such a quick response!!

PR #806, opened 2016-04-26
Viscosity only needs the OpenVPN folder + plist

### LINKED ISSUE #364: Viscosity syncs .Sparkle directory
Viscosity backups include the .Sparkle directory in Application Support. I've emailed the developer to ask them to move it but in the meantime, can we exclude it somehow? Viscosity was added in #45.

anthonysomerset:
Including the whole Application support folder includes the .Sparkle folder which seems to get very large over time with viscosity in particular as its not clearing out old versions like other apps seem to do

thankfully Viscosity only seems to create one folder in here so if we capture that then everything still fine

effectively fixes #364

### DIFF
--- a/mackup/applications/viscosity.cfg
+++ b/mackup/applications/viscosity.cfg
@@ -2,5 +2,5 @@
 name = Viscosity
 
 [configuration_files]
-Library/Application Support/Viscosity
+Library/Application Support/Viscosity/OpenVPN
 Library/Preferences/com.viscosityvpn.Viscosity.plist


lra:
thanks

Owpac:
Hey there!

Sorry to dig an old PR, but I don't think it is a good idea to add this folder `Library/Application Support/Viscosity/OpenVPN` to be saved since it contains private keys and certificates, it should not be upload somewhere else than on your local computer.

What do you think about it @anthonysomerset?

PR #808, opened 2016-04-28
[#autojump] Add support for autojump

welly:
A useful navigation utility https://github.com/wting/autojump

### DIFF
--- a/mackup/applications/autojump.cfg
+++ b/mackup/applications/autojump.cfg
@@ -2,4 +2,4 @@
 name = Autojump
 
 [configuration_files]
-Library/autojump
+Library/autojump/autojump.txt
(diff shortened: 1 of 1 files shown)


mackup/applications/autojump.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = Autojump
+
+[configuration_files]
+Library/autojump
```
lra: new line missing.
There are log files in this folder, please only sync config files.

lra:
Please follow https://github.com/lra/mackup/blob/master/CONTRIBUTING.md

welly:
I'm thinking now that this probably isn't quite suitable for mackup because the log file is the only thing that autojump generates or has in the way of config. I'll close this one.

PR #812, opened 2016-05-13
Adding Fisherman Support

sallar:
[Fisherman](http://fisherman.sh/) is a package manager for Fish Shell. This PR adds the support for that.


mackup/applications/fisherman.cfg
```diff
@@ -0,0 +1,6 @@
+[application]
+name = Fisherman
+
+[configuration_files]
+.config/fisherman
```
laughedelic: I believe that this folder shouldn't be synced because may contain git files. `~/.config/fish/fishfile` is all you need: you just run `fisherman` without subcommands and it installs all your packages from the fishfile.

@jbucaran correct me if I'm wrong.

sallar: @laughedelic You are right. Fixed. :)

sallar: @laughedelic You are right, but a bit late. Maybe I can open a new PR?

laughedelic: @sallar sure, it's not late to fix it 👍 

I spent quite some time just now, fighting with google-drive which tried to sync my local plugins `.git` folders 😅 it didn't want to unsync it, and it's thousands of files..

jbucaran: @sallar @laughedelic Thanks for the mention. ~/.config/fisherman is like `node_modules`, so it should definitely _not_ be synced. The fishfile is similar to `package.json`'s `pkg.dependencies`.

If you have installed fisherman, running:

```
fisher
```

in your terminal will cause fisherman to install everything listed in the fishfile.

sallar: I created a new PR :)

jbucaran:
@sallar Could you please edit https://github.com/lra/mackup/pull/812#issue-154678568 and replace the URL for https://fisherman.github.io

PR #817, opened 2016-05-18
Fix missing alternative in FileZilla entry

drbyte:


drbyte:
Agreed with @leek -- please merge #833 instead of this one  (after he updates the branch)

lra:
done, thanks

PR #822, opened 2016-05-26
Avoid backup on whole .docker folder (that could contain docker machi…

gdraynz:
…nes)


darcy:
👍

dr-impossible:
+1

matti:
`.docker/config.json` contains authentication tokens. Is this allowed or not? `CONTRIBUTING.MD` says `Do not sync sensitive information, like clear passwords or private keys`

I'd say that let people sync what they want, but what's the policy here?

matti:
Same goes to s3cmd. Are tokens considered private keys? @lra ?

dr-impossible:
The AWS secrecy key definitely is. The AWS console reveals it only when you
first create the key pair and then never again.

PR #823, opened 2016-05-27
ZSH: sync .zsh_history

kadel:
Add support for syncing .zsh_history file


jnv:
I am not even sure this is a common name for the history file. Some ZSH configurations use for example `.zhistory`. 

I have actually tried to sync my history file and ZSH always replaced the symlink with an actual file, so Mackup complained about an existing file. Not to mention you will probably get conflicted copies. But YMMV.
I think service like https://bashhub.com/ is more robust solution.

derimagia:
@jnv although I don't think history should be a default in this -  ".zhistory" is the default in zsh. ".zsh_history" is popular because oh-my-zsh sets it to that (https://github.com/robbyrussell/oh-my-zsh/blob/3705d47bb3f3229234cba992320eadc97a221caf/lib/history.zsh#L3)

PR #826, opened 2016-06-03
Changed the pip command to install system wide

iainbeeston:
The mac install instructions suggest using homebrew, which installs
mackup system wide in /usr/local. For other operating systems it
suggests pip, but by default pip installs to the user's home
directory. That's great, but it's not automatically in the $PATH (like
/usr/local/bin is) and this can be confusing for users who aren't used
to python.

I've changed the install instructions to recommend installing system
wide, because of this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,12 @@ If not running OS X, or you don't like Homebrew, you can use [pip](https://pip.p
 > is already installed on your system.
 > If this is the case, it will be upgraded to the latest version.
 
+> On **Ubuntu**, pip will install to the current user's home
+> directory rather than system-wide. Because of this, when
+> installing pip on **Ubuntu** you will need to run `pip install`
+> with the `--system` flag as well (on other platforms this is not
+> needed)
+
 ```bash
 # Install Mackup with PIP
 pip install --upgrade mackup


lra:
that's great, didn't know this, I think pip behavior changed.
But I can't get this to work on my mac?
```
[analogue:~] 3s $ pip install --system mackup

Usage:   
  pip install [options] <requirement specifier> [package-index-options] ...
  pip install [options] -r <requirements file> [package-index-options] ...
  pip install [options] [-e] <vcs project url> ...
  pip install [options] [-e] <local project path> ...
  pip install [options] <archive url/path> ...

no such option: --system
```

iainbeeston:
This is irritating, it seems the default for `pip install` is different on debian-based versions of linux. Here's what I get on ubuntu:

```
pip install --help | grep system
  --system                    Install using the system scheme (overrides
                              --user on Debian systems)
```

Seems like either the default is `--user` on debian, or at least the `--system` option only exists on debian. Perhaps there should be separate install instructions for debian/ubuntu?

lra:
> Perhaps there should be separate install instructions for debian/ubuntu?

Sure, can you change your PR?

iainbeeston:
Sorry for the delay, I've looked into this further now, and it seems that ubuntu actually patches pip to change the default to install user-locally (not globally), so it's only on ubuntu that users need to use the `--system` flag (see [here](https://bugs.launchpad.net/ubuntu/+source/python-pip/+bug/1460203))

I've updated the wording to reflect that. I think this should be ok to merge now?

alancwoo:
Wow I was really losing it trying to solve this problem for over an hour on Ubuntu 18.10, why isn't this merged into the docs, if at least a comment for Ubuntu?

lra:
Merging it now, this is annoying, but nice find!

PR #829, opened 2016-06-07
Update README.md to provide information on upgrade with PIP

orschiro:


README.md
```diff
@@ -41,6 +41,9 @@ If not running OS X, or you don't like Homebrew, you can use [pip](https://pip.p
 # Install Mackup with PIP
 pip install mackup
 
+# Update an existing version of Mackup with PIP
+pip install -upgrade mackup
```
lra: This would be `--upgrade`, but I suggest you specify that it's either `pip install` or `pip install --upgrade` and not both.

orschiro: Thanks! I think you're absolutely right. What about now?

README.md
```diff
@@ -39,7 +39,9 @@ If not running OS X, or you don't like Homebrew, you can use [pip](https://pip.p
 
 ```bash
 # Install Mackup with PIP
-pip install mackup
+pip install --upgrade mackup
+
+Note: If a previous version of Mackup already exists, it will be upgraded to the latest version
```
lra: It's not a command, it should be a comment. Also the comment should be before the commands it applies to. You don't want to know what a command does after you execute it.

lra:
Can you fix the markdown errors? See circleci below.

orschiro:
Dear @lra,

I really appreciate your feedback!

I hope it's better now. :smiley: 

~Robert

PR #835, opened 2016-06-18
allow an 'abspath' config setting to set an absolute storage path

shreeve:
When using Dropbox, the `path` is constructed for the user. If, however, the user has placed the Dropbox folder in another location, the path won't match what Mackup assumes. This `abspath` config option in the `storage` section allows an absolute path to be provided for the base directory for Mackup.

### DIFF
--- a/mackup/config.py
+++ b/mackup/config.py
@@ -218,6 +218,10 @@ class Config(object):
                 raise ConfigError("The required 'path' can't be found while"
                                   " the 'file_system' engine is used.")
 
+        # allow an absolute path override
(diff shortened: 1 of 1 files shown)


lra:
Not sure I understand the issue, Mackup is supposed to detect the Dropbox path automatically. If there is a bug let's fix it, but requiring the user to manually set a path is not a scalable solution IMO, as it would have to be done on every workstation.

shreeve:
@lra - I had to add this when I changed by top-level Dropbox folder location. Apparently, when I rebooted, this was no longer needed. So... I'll close this PR.

PR #837, opened 2016-06-20
Add application config file for Webstorm 2016.1

halhenke:
Basically exactly the same path/file structure as for previous versions of WebStorm.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## WIP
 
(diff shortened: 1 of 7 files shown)


purpleblues:
+1

lra:
Would you mind merging this and the other webstorm ini into 1?
Like https://github.com/lra/mackup/blob/master/mackup/applications/intellijidea.cfg

halhenke:
Ah gotcha - yep can do 👍

halhenke:
@lra done

PR #841, opened 2016-06-28
Updating Oh My Zsh config to be more specific

leek:
Syncing the whole `.oh-my-zsh` directory is unnecessary and was breaking for me. User's custom scripts/config should be located in the `custom` directory only - so let's just sync that.

### DIFF
--- a/mackup/applications/oh-my-zsh.cfg
+++ b/mackup/applications/oh-my-zsh.cfg
@@ -2,4 +2,4 @@
 name = Oh My Zsh
 
 [configuration_files]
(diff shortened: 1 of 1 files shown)


vitorgalvao:
This won’t suffice if you installed a custom theme.

leek:
**@vitorgalvao:** Custom themes should also be placed in the `custom` directory. See: [Overriding and adding themes](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization#overriding-and-adding-themes)

PR #844, opened 2016-07-06
Add support for mbsync.

tricon:
Adds support for mbsync: http://isync.sourceforge.net/mbsync.html


tricon:
I'll initiate this from a proper branch.

PR #846, opened 2016-07-12
Add support for Fantastical 2

bleen:


lra:
Can you instead merge the conf with the existing fantastical?

bleen:
I wasn't sure if it was acceptable to have files that may or may not exist listed in a cfg file... sounds like you'd be ok with that though?

lra:
Totally

bleen:
Cool ... Updated pr

PR #855, opened 2016-08-02
fix(config.py): replace 'str(path)' with 'path.encode(utf-8)'

wy-z:
I got UnicodeEncodeError when do backup with engine `google_drive`.
eg. `str(u"Google 云端硬盘")` 

Signed-off-by: weiyang weiyang.ones@gmail.com


ToostInc:
You might want to remove the `path.decode('utf-8')` line as well, since .decode() doesn't even exist in py3 anymore. It actually gives an AttributeError. I think the smaller than got mixed up ;)

Issue #784, opened 2016-03-21
Feature request: user friendly error message


loggerhead:
It's not intuitive for user to see error message like below:

```
Restoring .mackup.cfg ...
You already have a file named .mackup.cfg in your home.
Do you want to replace it with your backup ? <Yes|No>n
Traceback (most recent call last):
  File "/usr/local/Cellar/mackup/0.8.13/libexec/bin/mackup", line 9, in <module>
    load_entry_point('mackup==0.8.13', 'console_scripts', 'mackup')()
  File "/usr/local/Cellar/mackup/0.8.13/libexec/lib/python2.7/site-packages/mackup/main.py", line 121, in main
    app.restore()
  File "/usr/local/Cellar/mackup/0.8.13/libexec/lib/python2.7/site-packages/mackup/application.py", line 157, in restore
    home_filepath))
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/posixpath.py", line 156, in samefile
    s2 = os.stat(f2)
OSError: [Errno 2] No such file or directory: '/Users/fz/.ackrc'
```

Maybe change to more readable error message would be great help for newbies.

Issue #559, opened 2015-06-02
No such file or directory: ~/Dropbox/Mackup/.hgrc 


aristidesfl:
I get this message when I try to restore on a new machine:

```
$ mackup restore
Traceback (most recent call last):
  File "/usr/local/Cellar/mackup/0.8.5/libexec/bin/mackup", line 9, in <module>
    load_entry_point('mackup==0.8.5', 'console_scripts', 'mackup')()
  File "/usr/local/Cellar/mackup/0.8.5/libexec/lib/python2.7/site-packages/mackup/main.py", line 74, in main
    app.restore()
  File "/usr/local/Cellar/mackup/0.8.5/libexec/lib/python2.7/site-packages/mackup/application.py", line 125, in restore
    home_filepath))
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/posixpath.py", line 162, in samefile
    s1 = os.stat(f1)
OSError: [Errno 2] No such file or directory: '/Users/user/Dropbox/Mackup/.hgrc'

```

Facts:

1. I did not back up mercurial
2. Mercurial is **not** on my list of applications to backup
3. Therefore there is no .hgrc on my Dropbox
4. The new machine where I did restore, didn't have a .mackup config file, so it had to figure out which applications to restore
5. Backing up .mackup solves this problem because it won't try to restore mercurial

Questions:

1. How does mackup decide which files to restore when there is no black/white list?
2. Is Mackup checking which files already exist on the machine to decide which files to restore?
3. If so, shouldn't it instead check which files exist on the backup?

joeyhoer:
I'm seeing this exact same issue.

On a machine that already has applications installed, but where mackup has not yet run. I get the `No such file or directory` error for the  applications that are not in the back up (or are only installed on the "new" machine).

@aristidesfl, I agree that mackup should check to see if the files exist in the backup before trying to restore them.

kfinlay:
EThe
Sent using CloudMagic [https://cloudmagic.com/k/d/mailapp?ct=pi&cv=6.1.51&pv=8.3]
On Sat, Jun 13, 2015 at 10:53, Joey Hoer <notifications@github.com> wrote:
I'm seeing this exact same issue.

On a machine that already has applications installed, but where mackup has not
yet run. I get the No such file or directory error for the applications that are not in the back up (or are only installed
on the "new" machine).

@aristidesfl [https://github.com/aristidesfl] , I agree that mackup should check to see if the files exist in the backup
[...]

Issue #710, opened 2015-11-09
Restore Error


Anzumana:
Traceback (most recent call last):
      File "/usr/local/Cellar/mackup/0.8.12/libexec/bin/mackup", line 9, in <module>
        load_entry_point('mackup==0.8.12', 'console_scripts', 'mackup')()
      File "/usr/local/Cellar/mackup/0.8.12/libexec/lib/python2.7/site-packages/mackup/main.py", line 121, in main
        app.restore()
      File "/usr/local/Cellar/mackup/0.8.12/libexec/lib/python2.7/site-packages/mackup/application.py", line 157, in restore
    home_filepath))
      File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/posixpath.py", line 156, in samefile
        s2 = os.stat(f2)
    OSError: [Errno 2] No such file or directory:     '/Users/anzumana/Library/Preferences/com.googlecode.iterm2.plist'

Systeminformation:
iMac 10.10.5
Python 2.7.5
zsh

Not sure how i can debug to figure this one out.
Any help greatly appreciated.
Love the app btw :)

Hout:
Here is another one:
```
$ mackup restore
Traceback (most recent call last):
  File "/usr/local/Cellar/mackup/0.8.12/libexec/bin/mackup", line 9, in <module>
    load_entry_point('mackup==0.8.12', 'console_scripts', 'mackup')()
  File "/usr/local/Cellar/mackup/0.8.12/libexec/lib/python2.7/site-packages/mackup/main.py", line 121, in main
    app.restore()
  File "/usr/local/Cellar/mackup/0.8.12/libexec/lib/python2.7/site-packages/mackup/application.py", line 157, in restore
    home_filepath))
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/posixpath.py", line 156, in samefile
    s2 = os.stat(f2)
OSError: [Errno 2] No such file or directory: '/Users/jeroen/Library/Preferences/com.adobe.Photoshop.plist'
```
Python 2.7.10
OSX 10.11.4 beta
bash

Hout:
This happened after I changed my user name. There are still old symbolic links that point to the previous home directory that does not exist anymore.

Hout:
It would be a nice to have if mackup would remove links in the restore process IF the target does not exist anymore.

glaszig:
[...]

PR #863, opened 2016-08-14
Fix unexpected exception when the home file is a broken link and add related test cases.

zhangfand:
When `home_filepath` is a broken link, the `os.path.samefile` will raise exception `OSError: [Errno 2] No such file or directory`.The patchs fix this problem and also add related test cases.

It should also fix issues #784 and #559.


kmcnellis:
This also addresses #710

max-sixty:
I realize it's eight years later, but this would be great to merge if it still works — the workarounds such as https://github.com/lra/mackup/issues/559#issuecomment-114193636 are inconvenient...

Issue #449, opened 2015-01-27
TextMate 2 syncing seems broke ruby runtime


morphinewan:
I use [ctrl + cmd + o] to open a terminal frequently. But I found out that the ruby runtime it depends on will be broken after I synced TextMate 2 configuration files.

Is this some kind of a bug?

morphinewan:
When I run open terminal function at the first time, it will install a ruby runtime in to Library/Application Support/TextMate/Ruby.

It seems make Library/Application Support/TextMate a symbolic link will make TextMate 2 can't use its ruby installed inside properly.

graeson:
After installing mackup and running `mackup backup`, textmate throws the following error: "Failure running 'Comment Selection'. env: ruby 18: Permission Denied"

jrk:
+1 this is a huge problem for me. It breaks a very large number of core TextMate operations.

lra:
Should we remove `Library/Application Support/TextMate` from https://github.com/lra/mackup/blob/master/mackup/applications/textmate.cfg to not sync it?

nicolinuxfr:
I think I have the same problem… still. Maybe it's because I synced TextMate before the fix ?

But in the Mackup folder, I only saw bundles and not the whole TextMate folder, so I don't understand what happens…

joshS314159:
I'm encountering this same error with Textmate once I run backup: `"Failure running 'Comment Selection'. env: ruby 18: Permission Denied"`.

Does anyone have a solution?

nicolinuxfr:
No real solution, but here's what I do when it happens : edit the command, and change `ruby18` to just `ruby` at the end of first line.

For me, it was a working fix each time. And I have not seen any functional difference after the change.

joshS314159:
Thanks! That was a real easy fix.

jrk:
The issue seems to be that sync services aren't always preserving the execute permission bits on synced files. The more direct fix is:

```bash
$ cd Mackup/Library/Application Support/TextMate/Managed/Bundles/Bundle Support.tmbundle/Support/shared/bin
$ chmod +x *
```

nicolinuxfr:
Indeed, I can confirm it works.

So, is it a bad idea to use Mackup ? If it changes execute permission each times, it will be a nightmare…

miguelandres:
This is still happening, using google drive at least

Issue #249, opened 2014-04-12
TextMate "Favorites" are symlinks - Mackup copies entire path into Dropbox


paste:
TextMate stores its Favorites as symlinks within the `~/Library/Application Support/TextMate/Favorites` directory:
![screenshot 2014-04-12 03 48 13](https://cloud.githubusercontent.com/assets/136583/2687070/612e5460-c232-11e3-8dcb-fda8a5b29f8b.png)

When Mackup tried to backup my TextMate configuration, the symlinks were expanded so that 20GB worth of "Favorites" were copied into Dropbox/Mackup.

Also, it's silly that you can't use the `mackup` command to backup and restore individual apps. Why can't I just do: `mackup restore TextMate`?

lra:
Fixed in https://github.com/lra/mackup/commit/34b727be360f5c17396d5630ddeef12c55cb43e1

Issue #100, opened 2013-08-07
Crashed while backing up TextMate files


cfr:
Backing up Library/Application Support/TextMate...
    Traceback (most recent call last):
      File "/usr/local/bin/mackup", line 947, in <module>
        main()
      File "/usr/local/bin/mackup", line 900, in main
        app.backup()
      File "/usr/local/bin/mackup", line 449, in backup
        copy(filepath, mackup_filepath)
      File "/usr/local/bin/mackup", line 684, in copy
        shutil.copytree(src, dst)
      File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/shutil.py", line 205, in copytree
        raise Error, errors
    shutil.Error: [('/Users/me/Library/Application Support/TextMate/Managed/Bundles/iPhone.tmbundle/Support/objc_completion2.rb', '/Users/me/Dropbox/Mackup/Library/Application Support/TextMate/Managed/Bundles/iPhone.tmbundle/Support/objc_completion2.rb', "[Errno 2] No such file or directory: '/Users/me/Library/Application Support/TextMate/Managed/Bundles/iPhone.tmbundle/Support/objc_completion2.rb'"), ('/Users/me/Library/Application Support/TextMate/Managed/Bundles/iPhone.tmbundle/Support/SpecialRules.txt', '/Users/me/Dropbox/Mackup/Library/Application Support/TextMate/Managed/Bundles/iPhone.tmbundle/Support/SpecialRules.txt', "[Errno 2] No such file or directory: '/Users/me/Library/Application Support/TextMate/Managed/Bundles/iPhone.tmbundle/Support/SpecialRules.txt'")]

lra:
Do you still experience the problem?

lra:
Fixed in https://github.com/lra/mackup/commit/34b727be360f5c17396d5630ddeef12c55cb43e1

PR #897, opened 2016-10-11
Fix with including TextMate bundles path.

huacnlee:
I saw it removed by 34b727be360f5c17396d5630ddeef12c55cb43e1, please recover again.

Because in Bundles path, it may be have a lot of custom changes.
For example: Snippets, KeyMap changes.


lra:
A lot of people had problem with it: https://github.com/lra/mackup/issues/449 https://github.com/lra/mackup/issues/249 https://github.com/lra/mackup/issues/100

Did you try the config yourself?

lra:
oh it's only bundles, ok

PR #910, opened 2016-11-04
Add option to generate custom app config by name

alin23:
## Problem
Creating custom app configs by hand is slow.

## Solution
A single command to generate a config and populate it with relevant data based on the app name.
The algorithm searches the following locations for filenames that contain the app name:
* ~/Library/Preferences
* ~/.config
* ~/    (only filenames starting with `.`)

#### Example:
`mackup generate hyper`

will create the file _~/.mackup/hyper.cfg_ with the following content:

```ini
[application]
name = Hyper

[configuration_files]
.hyper.js
```


lra:
why closing it? not working?

alin23:
It was not getting any attention so I thought it may not be relevant to the project. Do you want me to reopen it?

PR #924, opened 2016-11-23
Add Boxer

icopp:
This adds the DOS game emulator [Boxer](http://boxerapp.com). It also includes the default folder the app keeps the packaged games in, since most of the time they'll be pretty small (5 MB or less).


mackup/applications/boxer.cfg
```diff
@@ -0,0 +1,6 @@
+[application]
+name = Boxer
+
+[configuration_files]
+Library/Preferences/net.washboardabs.boxer.plist
+Applications/DOS Games
```
lra: Not sure what you want to sync here.

1. Is there really an `~/Applications` folder being created?
1. Do we really want to sync whole games? It might mean a lot of big files no?

I feel like we should just sync settings.

icopp: > Is there really an `~/Applications` folder being created?

Yes.

> Do we really want to sync whole games? It might mean a lot of big files no?

These are almost never going to exceed maybe 15 MB at the very most for each game, so it seemed like it could be useful to include.

PR #926, opened 2016-11-23
Add global .gitignore to Git

icopp:
This improves the existing Git support by also accounting for a global `.gitignore` file.


aurimasniekis:
+1

lra:
I can't see any official documentation about a `~/.gitignore` file.

> Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by core.excludesFile in the user’s ~/.gitconfig. Its default value is $XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set or empty, $HOME/.config/git/ignore is used instead.

https://git-scm.com/docs/gitignore

PR #931, opened 2016-12-02
Thunderbird

orschiro:


mackup/applications/thunderbird.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = Thunderbird
+
+[configuration_files]
+.thunderbird
```
lra: Are you sure it's fine to sync the whole folder? Isn't there a lot of stuff in there, also local specific files and cache?

orschiro: Good point. I guess we should not sync the "Mail" folders. However, how to do that given that the profile folder contains a random name?

image

revolter: It should support regex like `.thunderbird/\w+.default`

orschiro: @revolter this would only sync the default folder?

revolter: It was just an example, I don't know what exactly is needed for Thunderbird. My points was that a separate PR should be made so mackup supports regex when matching instead of simple strings.

orschiro: @revolter thanks. I am going to close this PR for now as I don't know how to properly configure TB to work with Mackup. Currently also not using TB.

PR #952, opened 2017-01-03
Added user configuration of zsh located in .zshrc.local

jsilence:
Changes to the zsh configuration are often put into .zshrc.local which is then usually included via .zshrc.

Makes sense to backup and symlink this file as well.


lra:
This sounds like user config, I can't find an official reference of `.zshrc.local` in the zsh doc.
You might want to add it to your user config.

jsilence:
Well, it is recommended on http://grml.org:

´´´
# Optionally also grab the user configration:
# wget -O .zshrc.local  http://git.grml.org/f/grml-etc-core/etc/skel/.zshrc
´´´

In zsh forum discussions the grml zsh configuration is often recommended as an alternative to the oh-my-zsh setup. This made its way into some semi-official documentation like for example the german Arch Linux zsh page (https://wiki.archlinux.de/title/Zsh).

lra:
Yes but none is officially supported by ZSH, it's just distribution specific forks.
If we start accepting those, the conf files will be very subjectives and will have to include every habit and fork ever made.

I suggest using the user config to sync it until it becomes official.

jsilence:
Fair enough!

PR #955, opened 2017-01-04
Adding support for postico

iloveitaly:
Here's a config for Postico. The first time I tried to sync using this setup, I got some weird errors in Postico. Second time, everything worked fine.

It would be best if someone else verified this before merging.


lra:
Can you confirm that it's working across computers? Because store applications are in a sandbox and are not supposed to work =(

See https://github.com/lra/mackup/issues/615

PR #957, opened 2017-01-06
Update maid

dnicolson:
This pull request backs up only the `.maid/rules.rb` file, if the whole `.maid` directory is backed up then the `.maid/maid.log` file would be included as well.


vitorgalvao:
Not everyone has a `rules.rb`. Some people have multiple and name them different things. Hence why the whole directory makes more sense.

PR #971, opened 2017-02-09
don't try to chmod temporarily orphaned symlinks during restore

glaszig:
actual use case (as layed out in [#710](https://github.com/lra/mackup/issues/710#issuecomment-278811040))

let's say you have fish and fisherman backed up with mackup.
fisherman puts links into fish's config directory.

on a new machine you then run mackup restore
and because of the way mackup handles .mackup.cfg entries,
it might try to restore fish with its symlinks first
-- essentially restoring orphaned links -- and then
it restores fisherman and its files, making the symlinks valid again.

this patch modifies chmod tests to reflect the situation and fixes the crash i'm currently experiencing:
```
$ nosetests
..................E..............
======================================================================
ERROR: test_chmod_file (utils_test.TestMackup)
----------------------------------------------------------------------
  File "/Users/martin/Code/mackup/tests/utils_test.py", line 289, in test_chmod_file
    utils.chmod(dir_name)
  File "/Users/martin/Code/mackup/mackup/utils.py", line 178, in chmod
    os.chmod(os.path.join(root, cur_file), file_mode)
OSError: [Errno 2] No such file or directory: '/var/folders/r0/xh9rk2794rl1yndl3crn91x00000gn/T/mackup-tests-O9R3iC/link'
-------------------- >> begin captured stdout << ---------------------
--

--------------------- >> end captured stdout << ----------------------

----------------------------------------------------------------------
Ran 33 tests in 0.110s

FAILED (errors=1)
```

after: 
```
$ nosetests
.................................
----------------------------------------------------------------------
Ran 33 tests in 0.197s

OK
```

### DIFF
--- a/mackup/utils.py
+++ b/mackup/utils.py
@@ -175,7 +175,9 @@ def chmod(target):
             for cur_dir in dirs:
                 os.chmod(os.path.join(root, cur_dir), folder_mode)
             for cur_file in files:
-                os.chmod(os.path.join(root, cur_file), file_mode)
+                fullpath = os.path.join(root, cur_file)
+                if os.path.isfile(fullpath):
+                    os.chmod(fullpath, file_mode)
 
     else:
         raise ValueError("Unsupported file type: {}".format(target))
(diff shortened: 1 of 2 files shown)


glaszig:
hello?

lra:
Thanks for this — the bug is still valid (chmod still crashes on orphaned symlinks during restore). I've reworked the fix for the current `src/` layout and pytest in #2198, crediting your original work. Closing in favor of that.

PR #983, opened 2017-03-02
Remove Fisherman's Modules Folder

sallar:
As mentioned by @laughedelic (in #812), The modules folder doesn't need to be synced since everything will be installed again if the user runs `fisherman` command without arguments.


dyross:
Can we get this merged? Ran into some unwanted issues because of it. Is it possible to manually override this locally?

dyross:
Can we get this merged? Running into some issues because of it. Is there a way to override this locally?

Issue #869, opened 2016-08-24
.mackup.cfg [configuration_files] tag not being picked up


rkpatel33:
I have a .mackup.cfg file that looks like this:

![image](https://cloud.githubusercontent.com/assets/3799489/17950356/b970dd28-6a28-11e6-84f6-6ab4012bb7ff.png)

But when I run `mackup backup` the .eslint file does not seem to get picked up. I have confirmed the target file and config files are where they should be:

![image](https://cloud.githubusercontent.com/assets/3799489/17950415/1191f96a-6a29-11e6-8fe4-20b5bad29c99.png)

Any thoughts?

Rishi

acevif:
i have same problem

yurikoles:
Me too

revolter:
Me too. @lra, can I help you with this?

NewbMiao:
for now: 
step 1: 

```
cat .mackup/app-name.cfg
```
```
[application]
name = app-name

[configuration_files]
.some-config-file-name
.some-dir
```

step 2:

```
cat .mackup.cfg
```
```
[applications_to_sync]
app-name
```

then , `mackup backup` can work out!

revolter:
@lra, Shouldn't `mackup` backup the actual custom app config file created by the user that's used to sync configuration files? In @NewbMiao's example, the `.mackup/app-name.cfg` file.

mbaechtold:
One should omit step 2 or `app-name` will be the only app to be synched.

Synching your custom files is explained in https://github.com/lra/mackup/tree/master/doc#add-support-for-an-application-or-any-file-or-directory.

revolter:
So that's why Mackup wasn't working for me for the past almost half a year! Thanks a lot @mbaechtold.

@NewbMiao, please edit your answer so we won't have any more victims.

pikeas:
Just bit by this!

schemacs:
`configuration_files` only works for custom app cfg files. RotsiserMho/mackup#4 should also be ported to this repo.

lra:
Fixed with https://github.com/lra/mackup/pull/984

PR #984, opened 2017-03-03
Improve the docs about syncing custom files.

mbaechtold:
As reported in https://github.com/lra/mackup/issues/869, the `[configuration_files]` section of the `.mackup.cfg` is not working as advertised in the `Configuration` chapter of the docs.

The users should be pointed to the section where the syncing of custom apps (and individual files) is explained until #869 has been fixed.


lra (approved):
Thanks

PR #991, opened 2017-03-18
Added support for youtube-dl

volt4ire:


Kol9yN (approved):
Proper paths, good to use it

PR #999, opened 2017-04-11
fixed link for vscode insiders

mislavcimpersak:
The link for _Visual Studio Code Insiders_ was pointing to the same url as _Visual Studio Code_.

Since _insiders_ version has it's own url why not use that one 😃 

I'm not adding anything to changelog since this is more in the domain of fixing a typo in docs than some real substantial change.


mislavcimpersak:
Hmm, I have a feeling that something is not right with `codecov/project` check :suspect:

mislavcimpersak:
I've resolved conflicts, so this is ready for review and merge 🎉

PR #1013, opened 2017-05-15
Update support for PyCharm and add PyCharmCE.

scooby:
I'm guessing that Jetbrains will release 2017.2 and 2017.3, so it seems reasonable to simply add those entries now.

There's an open issue to add wildcards, I think that would be inappropriate as it could lead to some pretty horrific breakage and it's not much of a lift to update this every year or so.

Let me know if this doesn't work for you, I can restrict it to only adding 2017.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## WIP
 
 - Add support for Boxer (via @icopp)
+- Update support for Pycharm (via @scooby)
 
 ## Mackup 0.8.16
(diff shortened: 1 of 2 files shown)


MadeOfMagicAndWires:
Added @wangkezun to the credits as well for #1008. I'll keep this open a bit longer in case you wanted to change this.

lra:
Merged with https://github.com/lra/mackup/pull/2088

PR #1024, opened 2017-06-18
Remove note about OpenSSH private keys from README

njdancer:
Private keys were removed from OpenSSH config in dcb26ba.


mackup/main.py
```diff
@@ -22,13 +22,13 @@
  1. list: display a list of all supported applications.
  2. backup: sync your conf files to your synced storage, use this the 1st time
     you use Mackup. (Note that by default this will sync private keys used by
-    OpenSSH and GnuPG.)
+    and GnuPG.)
```
lra: by and ;)

njdancer: Lol can't believe I missed that.

njdancer: Fixed.

Issue #1012, opened 2017-05-13
Problem/Questions with BetterTouchTool


seanyhc:
I'm new to this so not sure if this is the expected behavior:
I wanted to sync BetterTouchTool, specifically my Chrome keyboard shortcuts (see attached image). But after setting up it did not sync. So I'm wondering whether mackup support backing up these details or just the basic preferences in BTT?

If it's suppose to sync, here are my steps and please advice:
- I backup on my personal computer, config it to use icloud, and mackup created files in the specified path
- go to my work computer, do mackup restore, it ask me to create a folder on dropbox, so I cancel
- copy the config file to work computer home directory, run restore again
- it say "You already have a file named .mackup.cfg in your home.
Do you want to replace it with your backup ? <Yes|No>"
- I reply yes and yes to all follow up questions for each app
- after it's done, the new shortcuts I created on my personal computer doesn' seem to be sync to my work computer

Thanks,
Sean
![blank_skitch_document](https://cloud.githubusercontent.com/assets/1082342/26028261/dedc13e4-37d1-11e7-8bf1-0127306bebf6.jpg)

TheSavior:
Looks like there is already a thread about this: https://github.com/lra/mackup/issues/520

PR #1028, opened 2017-07-03
Update BetterTouchTool configuration file paths

### LINKED ISSUE #520: Not working with BetterTouchTool (BTT)?
Am I the only one for whom this isn't working with BTT? The folders are being synced via Dropbox alright, but each of my machines appears to use a unique filename so the settings are not synced between them.

Example filename: BTT Settings 2015-04-18212ED135-AF39-4EF0-AC76-F8B3A09FFC3A

TheSavior:
This fixed #520 and #1012 for me.

@fifafu 

It looks like [this previous commit](https://github.com/lra/mackup/commit/c2d6e681fef915f35ab03561882d6ed647c16b8c) to the configuration file limited the files and changed it to a directory that no longer exists on recent versions of better touch tool. These files restored my entire configuration from my other computer, however, there *are* other files in that directory locally. Perhaps we should be backing up the entire folder again? 

screen shot 2017-07-03 at 4 35 52 pm

### DIFF
--- a/mackup/applications/bettertouchtool.cfg
+++ b/mackup/applications/bettertouchtool.cfg
@@ -3,4 +3,6 @@ name = BetterTouchTool
 
 [configuration_files]
 Library/Preferences/com.hegenberg.BetterTouchTool.plist
-Library/Application Support/BetterTouchTool/bttdata2
+Library/Application Support/BetterTouchTool/btt_data_store.v2
+Library/Application Support/BetterTouchTool/btt_data_store.v2-shm
+Library/Application Support/BetterTouchTool/btt_data_store.v2-wal


MadeOfMagicAndWires:
Also implemented by #1045 and #1089 please see if you can work out a triple credit system.

TheSavior:
It's okay, I don't need the credit. 😄

MadeOfMagicAndWires:
Alright, thanks for your work regardless!

fifafu:
Backing up the whole folder is definitely the way to go :-)
BTT will switch from the v2 filename suffixs to v3 with one of the next updates.

PR #1029, opened 2017-07-08
Transmission

HlubyLuk:
Add alternative configs path


julienbourdeau:
Can you explain the difference between those 2 paths?
```
/usr/local/var/transmission/blocklists
~/../../usr/local/var/transmission/blocklists
```

They probably go the the same directory or file but I think it's always best to use the absolute path.

HlubyLuk:
Problem was mackup do not working with absolut path

trace

```python
  File "/usr/local/bin/mackup", line 11, in <module>
    load_entry_point('mackup', 'console_scripts', 'mackup')()
  File "/Users/HlubyLuk/Workspace/python/mackup/mackup/main.py", line 65, in main
    app_db = ApplicationsDatabase()
  File "/Users/HlubyLuk/Workspace/python/mackup/mackup/appsdb.py", line 53, in __init__
    .format(path))
ValueError: Unsupported absolute path: /usr/local/var/transmission/blocklists
```

MadeOfMagicAndWires (dismissed):
Please don't add system files. Sync user-specfic configuration files only.

A normal user running mackup won't be able to restore these from a backup, and presumably the user-specific overwrite the system files anyway.

Also please revert the changes to .gitignore and [don't forget to add your changes to the CHANGELOG](https://github.com/lra/mackup/blob/master/CONTRIBUTING.md)

lra (approved):

PR #1031, opened 2017-07-18
Add the correct Exercism config file

JordanAdams:
It looks like the current implementation copies `.exercism`. However it would appear that the correct file is `.exercism.json`. You can see this in [the Exercism help page](http://exercism.io/help) or if you run `exercism configure`. I've also left in the previous definition as I suspect this perhaps where the config used to be stored.

image


JordanAdams:
@MadeOfMagicAndWires The change log should be up-to-date now :)

PR #1040, opened 2017-08-10
chattr: suppress some errors

doubleloop:
* chattr prints out error for example in case of symlink
("Operation not supported while reading flags")


lra:
Fixed #1515

PR #1041, opened 2017-08-16
Fasd support updated

doubleloop:
Added .fasdrc to configuration_files. 
This allows to walk around the issue noticed in
https://github.com/lra/mackup/pull/743#issuecomment-221783525
You can simply set 
```
FASD_DATA="$HOME/.fasd/fasd"
``` 
in .fasdrc and mackup will link the whole directory, so even if fasd deletes/creates data file, it will not break the symlink.


doubleloop:
@lra Any reason why this was not merged? Just rebased.

lra (approved):
LGTM

PR #1050, opened 2017-08-31
Update utils.py

lokendrarawat143:
added "--force-no" option for forcing 'no' to each question by default.


fharper:
It would be nice to merge this PR with #1050 and merge this PR to main @r4w4t . If you don't have time or can't, I'm happy to create another PR with these changes as this option would be great :)

lra:
Thanks!

Issue #1010, opened 2017-05-08
Support for Visual Studio for Mac


Dids:
Since it's based on MonoDevelop, the two main configuration files (as of writing this) are available at:
```sh
~/Library/Preferences/VisualStudio/7.0/EditingLayout.xml
~/Library/Preferences/VisualStudio/7.0/MonoDevelopProperties.xml
```

If mackup supports wildcards, then I guess `~/Library/Preferences/VisualStudio/**` would be better.

PR #1071, opened 2017-10-04
Support for Visual Studio for Mac

stephen2m:
Update config to backup installed extensions (update for #1010 )


ivmirx:
Are the name and #1010 reference correct? Because I see VS Code in the configs and not VS for Mac which is a different product.

stephen2m:
@ivmirx you're actually right....it's indeed VS Code!

ivmirx:
@stephen2m thanks, just double-checking if I need a PR for VS4Mac. :)

CHANGELOG.md
```diff
@@ -2,10 +2,10 @@
 
 ## WIP
 
-- Add support for Micro (via @icopp)
```
lra: Why removing this?

I'll add it back in master.

chuckd:
For anyone confused like I was: this was reverted in June, see #1170.

PR #982, opened 2017-03-01
Add additional Spectacle library path

iansym:
There is one more path for the Spectacle.app where the Shortcuts.json file is stored.
This PR adds the `Library/Application Support/Spectacle` directory to the config.

PR #1076, opened 2017-10-13
Update spectacle.cfg

knoopx:
Also sync shortcuts

### DIFF
--- a/mackup/applications/spectacle.cfg
+++ b/mackup/applications/spectacle.cfg
@@ -3,3 +3,4 @@ name = Spectacle
 
 [configuration_files]
 Library/Preferences/com.divisiblebyzero.Spectacle.plist
+Library/Application Support/Spectacle/Shortcuts.json


MadeOfMagicAndWires:
Also, turns out this was already worked on: #982

notcome:
@MadeOfMagicAndWires The pull request you referred to was blocked. @mbaechtold said that pull request did not update `CHANGELOG.md`, though I believe the pull request already did that. Could we simply pass the first pull request? This is just two line without any significant work required, but causing Spectacle unable to synchronize is really disturbing.

lra:
I just merged #982.
If you think only the `Shortcuts.json` file and not the whole folder should be synced, please let us know and we'll merge this one too.

notcome:
@lra Thanks for the merge! I think they are functionally equivalent since the only file in that directory is `Shortcuts.json`. I would say we can safely close this one, but I am not the author of this pull request, so that is just a suggestion.

lra:
Ok let's close this then, and revisit if someone has a better option.

PR #1085, opened 2017-11-01
T ruby gem

paxperscientiam:
title


mackup/applications/t.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = t
+
+[configuration_files]
+.trc
```
vitorgalvao: It has a “no newline at end of file” warning. Those are [actually meaningful](http://unix.stackexchange.com/a/18789).

paxperscientiam: Thanks, fixed. Did the same for the other two PRs.

paxperscientiam:
@MadeOfMagicAndWires Fixed, i think

PR #1089, opened 2017-11-08
Update bettertouchtool.cfg

foray1010:


foray1010:
@MadeOfMagicAndWires thanks for the review, updated

MadeOfMagicAndWires:
Cheers, circleci seems a bit busted but we'll merge as soon as that is fixed

PR #1091, opened 2017-11-09
Android Studio 3.0

ned1988:


lra:
You can open it and merge it into https://github.com/lra/mackup/blob/master/mackup/applications/androidstudio-13.cfg so we have one file for all the versions of androidstudio

pvinis:
@MadeOfMagicAndWires if the person has no Linux access, why is this pr rejected? just put a comment or a table for which app has support for macos and Linux.
@lra makes sense.

PR #1093, opened 2017-11-23
fix spelling

janpio:


lra:
Why closing? Makes sense.

janpio:
Because #1094 did it better ;)

PR #1101, opened 2017-12-08
Also sync Sublime Text 3 license file

itamaro:


lra:
You think it's a good idea to sync those? This is kindof sensible.

itamaro:
> You think it's a good idea to sync those? This is kindof sensible.

My thinking is - if I reinstall my Mac, then my Sublime installation would be significantly lacking if the license isn't there.

I guess it's a little sensitive (that's what you meant, right?), but I assume my sync'ed config is sensitive and the sync-engine isn't public.

Also, I saw that it sync'ed my AWS credentials, which is way more sensitive...

MadeOfMagicAndWires:
According to [this](https://forum.sublimetext.com/t/lost-license-key/7810) and [this](http://www.sublimetext.com/docs/3/portable_license_keys.html) the license files, when binary, will only work on the machines they've been entered on and portable license files can only be created on Windows, using the portable version of Sublime.

You can also retrieve licenses [here](https://www.sublimetext.com/retrieve_key) (As far as I know without cost, I'll admit I don't use Sublime).

Seeing how it either won't work or is indeed sensitive information, we should probably keep things as they are.

PR #1106, opened 2017-12-30
Change finding logic for Dropxbox folder paths

### LINKED ISSUE #301: Multiples Dropbox pathes
Hello,

As Dropbox introduced multiple boxes now (with both perso and pro) and Mackup doesn't use Dropbox API, there is a sync problem. 

The fix could be made replacing the use of `hosts.db` with `info.json`. The first time  Mackup could ask (if multiple path) for the good one.

Xeite:
Change Dropbox metadata file to info.json because newer Dropbox find
Dropbox folder paths using info.json instead of host.db. Also, there
are chances to have more than one Dropbox accounts, so support to
choose right one.

Resolve #301


MadeOfMagicAndWires:
[Please add changes and your name to the WIP section in CHANGELOG.md](https://github.com/lra/mackup/blob/master/CONTRIBUTING.md)

Also, do you know if any old versions still use host.db? 

Seeing how the issue is from 2014 I assume we don't have to worry about this but if older versions haven't been deprecated we might still need a fall-back to host.db. 

I couldn't find anything but it's best to be on the safe side.

Xeite:
Thanks for review.

Then do I have to add fallback logic for host.db?

MadeOfMagicAndWires:
Sorry for the late reply: Please do.

PR #1111, opened 2018-01-12
add Plover support

ohAitch:


ohAitch:
Oh yeah thanks
I'm guessing I can't just say `$APPDIRS_USER_DATA/plover`?

ohAitch:
Hm actually, is there any good way to bridge the configs between the OS's? The layout of the Mac and Linux config directory is identical, and having mackup maintain two unsynced copies of those would be weird

MadeOfMagicAndWires:
You don't need to worry about that, the files will be saved into the same directory regardless of origin.

As for the configuration files on linux, apparently it's [$HOME/.local/share/plover](https://github.com/openstenoproject/plover/issues/674) for the moment

Issue #859, opened 2016-08-06
Add a flag for not linking automatically


jymoulin:
Hi,

The default usage (make a link to a 'remote' directory) is not a backup but a synchronization tool. If you change your original file, as this is only a link, you also change the backup file. You cannot recover it.

It can be nice to add a flag to choose between a link or a copy, to provide a real backup.

Thanks !
Best,
jym

xkef:
I don't see why a simple script like below does not achieve the  thing (then rsync back would do it imho):

```
#!/bin/bash

# make backupdir
mkdir backup_$(date +%Y_%m_%d)
cd backup_$(date +%Y_%m_%d)

get_mackup_resources() {
  git clone https://github.com/lra/mackup
  mv mackup/mackup/applications .
  rm -fr mackup 
  for d in applications/*.cfg ; do
    cat $d | sed -r '/^\s*$/d' | \
             sed -r '/^\[/d' | \
             sed -r '/^name/d' | \
             sed -r '/^#/d' >> apps.txt
  done
  rm -fr applications
  cat apps.txt | sort > apps_sorted.txt
}

# copy everything to backupdir
get_mackup_resources
mkdir mackup
rsync -aLkru ~/ \
  --files-from=apps_xkef_sorted.txt \
  --ignore-errors \
  mackup/
```

KevinVlaanderen:
Morning,

Like the original poster, I'd also be interested in a way to copy instead of link. To make the case, I'll outline my setup.

My primary use case is backing up instead of syncing, for the purpose of facilitating a system restore in case I move to a new system or something happens to my machine.

My approach to this use case is that I want to provision my machine from original sources as much as possible, in an automated fashion as much as possible. Applications should be installed through their normal channels (e.g. brew), repos should be cloned from their origins (e.g. Github), data should be available through e.g. Dropbox. And for configuration files, Mackup fills in the gap.

I currently have a task running every day that, amongst other things, does the following:
- Dump all installed brew applications to a file
- Run mackup to backup any new application config files
- Push everything to a private git repository

Mackup's approach, with per-application whitelists, is a good fit with the above. However, linking iso copying is rather intrusive, as it limits my ability to move things around. Instead, I would prefer a side-effect free mechanism while still employing the whitelisting mechanism.

I'd be happy to discuss this further, and if the discussion leads to a solid idea of how this might work and what the impact would be, I am very willing to implement this and send in a PR.

lra:
[...]

PR #1117, opened 2018-01-17
Feature mode copy and code refactoring

urban-1:
Hi @lra / all,

This is an attempt to introduce `copy` mode in Mackup. However, while going through the code I could not resist restructuring things a bit... So I will explain this PR in two main sub-sections:

## Code Refactoring

The changes are mainly focused on defining clear interfaces between components and thus code has been moved around but not changed a lot. The most notable changes are:

-   Most of Mackup functionality should be provided by the `Mackup` class which serves as an entry point to mackup api and acts as a controller for individual commands. `Mackup` now holds:
    -   The global configuration (as before)
    -   The runtime parameters
    -   The application database
    -   The user's environment
-   The `ApplicationsDatabase` is now loading `ApplicationProfile` instances instead of `dict`s. Helper methods have been removed and replaced by `__getitem__` implementation which allows `[<appname>]`. This way the `Mackup` instance (and previously the `main.py`) does not need to loop over the DB and create profiles, it just loops and applies the operation requested.
-   The `ApplicationProfile` is mostly the same. One difference is that instead of passing individual config or runtime values (like `dry_run`) down to its methods, we now use the `Mackup` instance (which contains all of those). Although a minor modification, allows for more attributes to be added into `Mackup` (and thus the cmd line args or global config) without modifying the `ApplicationProfile` method signatures...

I wanna believe that most of the refactoring makes senses and makes the code easier to extend and maintain.

## Copy Mode

The implementation of the copy mode is a *global* one as opposed to a per-INI/app. This simplifies the code, keeps config-file changes to minimum and avoids few issues that need resolving. These include:

-   How do we deal with apps that can be both linked and copied? This is a user preference which as you said should be stored in the INI... but doing so makes that INI file not generic any more (since it contains a user preference).
-   One could introduce a section in the global config that dictates which apps should be backed up via copy instead of linking. This will complicate the logic in the `ApplicationProfile`, especially given that a user might change configuration of an already backed up app! The implementation of this tho with the new code is pretty straight forward (ApplicationProfile.backup can override mode)
-   I cannot come up with a valid case where one needs some apps linked and some apps copied but (over)thinking about it, this can be extended down to a per-file manner!!!

Due to the above I believe that a global configuration option is a fast win and the safest way to go. We will need to think of the reconfiguration problem and how we detect it (Mackup metadata file in storage, including version, mode, etc?)

## Merging

I understand that such code changes are risky, especially given the nature of this app. If you review the code and decide that we should proceed with a merge, I would suggest that a new branch is created on the upstream repo (lets say `rc-0.9.0`):

-   We merge my `feature_copy_mode` to `rc-0.9.0`
-   New application profiles will always be added to the old upstream `0.8.x`
-   `0.8.x` will only receive bug fixes (if any)
-   All new features go to `rc-0.9.0`
-   Devs and testers work with `rc-0.9.0`
-   Once `rc-0.9.0` is deemed stable, we merge it to `master`

Some conflicts are to be expected on any code changes in `0.8.x` but these shouldn't be too many.

## Notes:

-   Still WIP but some experimentation with `mackup show git` feature is included. This breaks atm the clean API and needs to be defined better and moved into `Mackup` class...

Let me know what you think and if we should proceed with merging.

Cheers,

Andreas


MadeOfMagicAndWires:
To be clear, I haven't looked at the code yet, as it's a bit much for right now, but some points missing from the overview:

So, just for clarity, what does `copy` mode *do* :smirk:? Is it meant to replace symlinking in `backup` as suggested in #859 or is it just used as a one time command to copy files?

Second, how does it handle filename conflicts? Say there's an older copy in the backup folder already
and I'm now running `copy` again. Say I've symlinked the files already and now decide I'd rather want copies?

urban-1:
The copy mode replaces the current linking mode. This is enabled by adding the following in `~/.mackup.cfg`:

```
[general]
mode = copy
```

The commands remain the same, with exception the `mackup uninstall` which is a dummy one and has no effect in this mode (just prints a message): 

- `mackup backup` will load all application profiles and copy every file listed  to the backup location. 
- Similarly `mackup restore` will copy from the backup location into the home directory overwriting any existing files. `restore` will warn if a file already exist and will prompt the user (displaying relevant date information).

Copy operations are implemented with [distutils.dir_util.copy_tree(src, dst, update=1)](https://docs.python.org/2/distutils/apiref.html#distutils.dir_util.copy_tree):

> If update is true, src will only be copied if dst does not exist, or if dst does exist but is older than src.

At the moment **there is no detection on if the user changed configuration from `link` to `copy`**. This is something that I can easily add (when I find some time) in order to avoid disasters... However, I would like to see if you are willing to consider merging before adding final touches :) (I am using it in copy mode on my boxes atm)

My time is limited but I will do my best to reply on this thread. I tend to have more time in weekends for coding/fixing. Let me know if you require more information.

MadeOfMagicAndWires:
Thanks for the quick reply. Seeing how intensive this PR is I think I'll leave the decision to @lra but good work nonetheless

xkef:
As mentioned in #859 how is this better than directly using rsync?

xkef:
ty @urban-1 for your insights. 

I guess the only value (except for the nicely written python code) in this repo is the applications db. However, it still is not handling my standard configs of vscode insider versions or non-std dotfiles paths well - it is very opinionated in this respect. I don't even want to mention handling complex zsh/vim configs with deep symbolic / hardlinks...

I will stick with my ansible orchestration combined with rsync and git for synchronization.

lra:
Copy mode was implemented in https://github.com/lra/mackup/releases/tag/0.9.0

PR #1135, opened 2018-03-10
Add support for Visual Studio for Mac (#1010)

ivmirx:
Hi, adding VS4Mac. It was previously called Xamarin Studio and got both the name and config paths changed after acquisition by Microsoft.

### DIFF
--- a/mackup/applications/vs4mac.cfg
+++ b/mackup/applications/vs4mac.cfg
@@ -1,0 +1,7 @@
+[application]
+name = Visual Studio for Mac
+
+[configuration_files]
+Library/VisualStudio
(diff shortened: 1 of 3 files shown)


mackup/applications/vs4mac.cfg
```diff
@@ -0,0 +1,7 @@
+[application]
+name = Visual Studio for Mac
+
+[configuration_files]
+Library/VisualStudio
```
lra: Which files are saved in this folder? Just config or binary stuff?

ivmirx: Just configs, really similar to Xamarin Studio: color themes, key bindings, style policies, snippets. All JSON and XML.

PR #1147, opened 2018-04-03
feat: only sync init.vim for neovim

foray1010:
the original approach that I did 2 years ago sync the whole `.config/nvim` which is unnecessary since `init.vim` includes all the information like which packages to install and its config
syncing the whole `.config/nvim` will result in uploading all plugins and their `.git` to the cloud which is thousands of files, and very easy to cause file conflicts while syncing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Support for RubyMine 2017.03 (via @kibitan)
 - Add additional Spectacle library path (via @iansym)
+- Only sync init.vim for neovim (via @foray1010)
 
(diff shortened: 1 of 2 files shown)


lra:
Thanks for improving it!

revolter:
Shouldn't it sync `colors` and `plugin` too?

foray1010:
@revolter I didn't use this two directories so maybe u could open a pr. `plugin` sounds like uploading whole plugins. If yes then I would suggest not syncing it, it is a nightmare to sync large files, especially the `.git` directory and dependencies, could take hours

PR #1157, opened 2018-04-27
Updated phpstorm.cfg to support version 2018.1

hellodeloo:


bleen:
This should be closed in favor of #1153

lra:
Dupe of #1153

PR #1159, opened 2018-05-02
convert dropbox path to `str` if python3

nguyenkims:
This one fixes the error `Error: Unable to find the storage folder` that happens with the following scenario:
- mackup is first installed using Python2 (via `brew` for example)
- mackup is uninstalled
- mackup is installed using Python 3 (via `pip` for example)

In this case, the storage folder stored in `.dropbox/host.db` is `bytes` and not `str` as Python3 expects.


lra:
The python 3 compatibility has been added in #1209. But feel free to modify this PR to add the test in master.

.gitignore
```diff
@@ -8,5 +8,8 @@
 # Generated by make test
 /.coverage
 
+.idea/
```
lra: this should be added to your own ignore file, not Mackup's

mackup/utils.py
```diff
@@ -205,6 +205,10 @@ def get_dropbox_folder_location():
         error("Unable to find your Dropbox install =(")
     dropbox_home = base64.b64decode(data[1]).decode()
 
+    # Need to convert to str if Python3
```
lra: not needed anymore

PR #1162, opened 2018-05-09
Add support for PhpStorm 2018.1

bleen:
I wish there was a way to do this more generically to support all versions of PHPStorm moving forward, but alas...


bleen:
closing this in favor of #1153

PR #1163, opened 2018-05-14
Allow bootstrapping from iCloud Drive

njdancer:
This update adds some logic to check iCloud Drive for a mackup config before defaulting to Dropbox. If found the default engine is set to iCloud allowing .mackup.cfg to be automatically linked. Otherwise Dropbox continues to be the default.


lucasff:
Why this wasn't merged yet @lra?

PR #1164, opened 2018-05-14
Add support for TablePlus

rizkysyazuli:
Added support for [TablePlus](https://tableplus.io), the database manager. This will backup the user's saved connection, [minus the password](https://github.com/TablePlus/TablePlus/issues/193#issuecomment-388692099), which is saved in Keychain.


mackup/applications/cyberduck.cfg
```diff
@@ -2,5 +2,8 @@
 name = Cyberduck
 
 [configuration_files]
+# based on https://trac.cyberduck.io/wiki/help/en/faq#Preferencesandapplicationsupportfileslocation
 Library/Application Support/Cyberduck
+Library/Containers/ch.sudo.cyberduck/Data/Library/Application Support/Cyberduck
+Library/Group Containers/G69SCX94XU.duck/Library/Application Support/duck
```
lra: I don't think this is gonna work, it's specific to each computer no?

rizkysyazuli: Yes. The `mackup` backup command works fine. i have it in my personal `.mackup` config folder. But i haven't tried running the `uninstall` and `restore` command.

But i just noticed the path in line 7 doesn't exists. If i'm not mistaken, that was for an older version of Cyberduck.

    Library/Containers/ch.sudo.cyberduck/Data/Library/Application Support/Cyberduck

But sure. i will try to split those PR a bit later. sorry for the trouble.

vyuldashev:
Any progress on this PR? Really missing TablePlus support.

mackup/applications/tableplus.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = TablePlus
+
+[configuration_files]
+Library/Application Support/com.tinyapp.TablePlus
```
LukeAskew: You also need `Library/Application Support/com.tinyapp.TablePlus/Data` in order to save the connections.

lra: subfolders are contained in their parent folder.

PR #1169, opened 2018-05-27
Add support for Tower 3

diego-vieira:
I just copied the config from Tower 2 to 3, however, I have not found a way to test as there are no instructions to compile the app.


lra:
Testing a new config is explained here: https://github.com/lra/mackup/tree/master/doc#add-support-for-an-application-or-any-file-or-directory

PR #1171, opened 2018-06-07
Update x11.cfg

paxperscientiam:


lra:
Why would you want to share your `.Xauthority` on multiple computers?

PR #1172, opened 2018-06-07
Gimp

paxperscientiam:
I need some input with this one.

So, there is a "default" gimp folder created by the application itself that lives in `library/application support/` on macos.

However, that folder contains subfolders for each major dot releases. EG: `library/application support/gimp/2.8/`

There is also the `.gimp/`.

Questions:
*  How to deal with config tree; does mackup support wildcards? EG: `library/application support/gimp/2.*/`
* What about locations in linux?
* Am I choosing all files that match constraints?

Thanks!


lra:
> How to deal with config tree; does mackup support wildcards? EG: library/application support/gimp/2.*/

The only supported way is to mention the parent folder, e.g. `library/application support/gimp`

> What about locations in linux?

Just add them in the same file, Mackup ignores locations that don't exist.

> Am I choosing all files that match constraints?

Not sure I understand.

paxperscientiam:
Hi @lra 

What I meant is this: should the manifest file only reference RC files or would it be appropriate to reference the entire ".gimp" directory?

In addition to RC files, .also includes the following sub-directories:
```text
brushes
curves
dynamics
environ
fonts
fractalexplorer
gfig
gflare
gimpressionist
gradients
interpreters
levels
modules
palettes
patterns
plug-ins
scripts
tags.xml
templates
themes
tmp
tool-options
tool-presets
```

fharper:
@paxperscientiam: any plans to update this PR and merge it?

I wanted to take a look at GIMP, but I saw there are two PRs never merged, so I want to check before.

Issue #1090, opened 2017-11-09
UnicodeEncodeError when running mackup restore or uninstall


ppptok:
Getting the following error when I attempt to run mackup restore or uninstall. I'm not sure how to troubleshoot this further from my end.

```Traceback (most recent call last):
  File "/usr/local/Cellar/mackup/0.8.16/libexec/bin/mackup", line 9, in <module>
    load_entry_point('mackup==0.8.16', 'console_scripts', 'mackup')()
  File "/usr/local/Cellar/mackup/0.8.16/libexec/lib/python2.7/site-packages/mackup/main.py", line 65, in main
    app_db = ApplicationsDatabase()
  File "/usr/local/Cellar/mackup/0.8.16/libexec/lib/python2.7/site-packages/mackup/appsdb.py", line 34, in __init__
    if config.read(config_file):
  File "/Library/Python/2.7/site-packages/backports/configparser/__init__.py", line 705, in read
    self._read(fp, filename)
  File "/Library/Python/2.7/site-packages/backports/configparser/__init__.py", line 1109, in _read
    optname = self.optionxform(optname.rstrip())
UnicodeEncodeError: 'ascii' codec can't encode character u'\xdc' in position 28: ordinal not in range(128)```

silvb:
I also get a `UnicodeEncodeError` when running `mackup backup` right after installation.

`
mackup backup
Traceback (most recent call last):
  File "/usr/local/Cellar/mackup/0.8.16/libexec/bin/mackup", line 9, in <module>
    load_entry_point('mackup==0.8.16', 'console_scripts', 'mackup')()
  File "/usr/local/Cellar/mackup/0.8.16/libexec/lib/python2.7/site-packages/mackup/main.py", line 65, in main
    app_db = ApplicationsDatabase()
  File "/usr/local/Cellar/mackup/0.8.16/libexec/lib/python2.7/site-packages/mackup/appsdb.py", line 34, in __init__
    if config.read(config_file):
  File "/Library/Python/2.7/site-packages/backports/configparser/__init__.py", line 705, in read
    self._read(fp, filename)
  File "/Library/Python/2.7/site-packages/backports/configparser/__init__.py", line 1109, in _read
    optname = self.optionxform(optname.rstrip())
UnicodeEncodeError: 'ascii' codec can't encode character u'\xdc' in position 28: ordinal not in range(128)
`

I created a config file `.mackup.cfg` in my home directory with the following content

`
[storage]
engine = file_system
[...]

PR #1182, opened 2018-07-23
Temporary fix UnicodeEncodeError

CatTail:
Related to #1090


lra:
Thanks, I think we should solve #1090 instead.

PR #1183, opened 2018-07-30
Support ESLint global configs.

sh78:
[ESLint configuration][1] can be set per project or globally with a file
in `$HOME`. There are five potential file names. ESLint can also be
configured by rules in a `~/package.json` file, but that is a different
system and therefore not included here.

[1]: https://eslint.org/docs/user-guide/configuring#configuring-eslint


lra:
> ESLint can also be configured by rules in a ~/package.json file, but that is a different system and therefore not included here.

Nice.

sh78:
Oops! Docs updated.

Thanks @lra !

PR #1189, opened 2018-07-30
Support latest qutebrowser version config paths.

sh78:
qutebrowser has [changed the configuration paths][1]. The schema is now:

```
macOS/OS X
.qutebrowser/config.py

Linux
.config/qutebrowser/config.py

Legacy
.config/qutebrowser/qutebrowser.conf
```

This commit includes all three, just to be on the safe side.

[1]: https://github.com/qutebrowser/qutebrowser/blob/master/doc/help/configuring.asciidoc#configuring-qutebrowser


lra:
Thanks for all those clean PRs!

PR #1200, opened 2018-08-20
Add global gitignores

SamuelNitsche:


lra:
The thing with those global git ignores is that those are not official (e.g. referenced in `git`'s source code).

Because of this, you should use a custom config to sync them, or we'll end up having to specify everyone's config in the official application configs.

SamuelNitsche:
Thanks, I didn't know this. I have now created a separate file for the synchronization.

lra:
Well, I don't know if I an accept this, "Gitignore" is an application. what I was saying is that you can use this file yourself by adding it into the `.mackup` folder per https://github.com/lra/mackup/tree/master/doc#add-support-for-an-application-or-any-file-or-directory

SamuelNitsche:
Oh, now I see what you mean. I'm fine with this :) Thanks for your feedback!

PR #1206, opened 2018-08-24
add application Typora

allenlsy:
Add support for Typora.

Backup themes


lra:
Can you add a line to the changelog?

allenlsy:
@lra Sure. I've just noted the format of the changelog

PR #1207, opened 2018-08-24
feat(oh-my-zsh): add completions default folder

H--o-l:
`oh-my-zsh` default folder for custom auto-completions is `.oh-my-zsh/completions` so I propose to add it.


lra:
Can you add a line to the changelog?

PR #1220, opened 2018-09-15
Add Idea2018.2 CE & UE config

beerChili:
For reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544519-Directories-used-by-the-IDE-to-store-settings-caches-plugins-and-logs


xohzzwn6kcj9:
waiting for this to be released. :)

PR #1221, opened 2018-09-16
Sync entire karabiner config directory

revolter:


lra:
Looks like it's an XDG config, so it should be something like:
```
[xdg_configuration_files]
karabiner
```

captnswing:
Can be closed. 
Ticket refers to Karabiner-Elements. 

Current behaviour of https://github.com/lra/mackup/blob/master/mackup/applications/karabiner-elements.cfg is backing up all of `~/.config/karabiner`

revolter:
So exactly what my PR did, right?

Issue #1222, opened 2018-09-19
Add support for xonsh shell


kmcm0:
configuration exists in a single rc file:  .xonshrc
https://xon.sh

PR #1223, opened 2018-09-19
Add support for Xonsh shell

kmcm0:
added: xonsh.cfg
modified:   CHANGELOG.md
modified:   README.md

Adding application xonsh, issue #1222

### DIFF
--- a/mackup/applications/xonsh.cfg
+++ b/mackup/applications/xonsh.cfg
@@ -1,5 +1,5 @@
 [application]
-name = xonsh
+name = Xonsh
 
 [configuration_files]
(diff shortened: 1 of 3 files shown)


mackup/applications/xonsh.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = xonsh
```
lra: Can you name it `Xonsh` in here? It's supposed to be the pretty name.

kmcm0: roger.

PR #970, opened 2017-02-08
Add Slack

sscotth:

PR #1241, opened 2018-10-27
Add support for Slack

ttuygun:


mackup/applications/slack.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = Slack
+
+[configuration_files]
+Library/Application Support/Slack
```
lra: I can see lots of local specific stuff in this folder, why syncing it?
screen shot 2018-10-29 at 1 28 32 pm

ttuygun: Caches can be ignored. It's because syncing signed-in slack accounts.

ionlights: Do you know how Slack keeps authorization credentials stored?

Just off intuition, this feels like a security risk, in the event someone decides to put this as part of a public repository. (I haven't looked into this at all.)

Also, this doesn't support Linux. 😕 (As far as I understand, `mackup`, is attempting to build for both? @lra)

lra: I'm not worried about the security, if people make their backup public I can't help them.

Re. GNU/Linux, that's true, but lots of applications only support one OS and support the other later, that's not a blocker.

The blocker here is that we should not sync non-configuration stuff, or cache.

ionlights: So for Linux, I meant that the path listed seems to only be the MacOS location, not a Linux location. (I don't have a present GUI install of Linux so I can't provide much feedback there.)

As for security, sure – but I definitely think there should be warnings if there aren't already. (I've never seen such warnings, so I'm not sure they exist.)

lra: There was a warning for private keys, it was ignored anyway. There is now an explanation of what it does https://github.com/lra/mackup/blob/master/mackup/main.py#L30-L34

But really, you have to explicitly configure the backup method to use (or it uses Dropbox, which is private by default) and you have to explicitly state the application you want to backup.

Re. Paths, I was talking about the same thing, applications supported in Mackup generally only start with the MacOS path, and when someone using Linux shows up, it adds the Linux path later, no harm in this.

jamiew:
@ttuygun per the other thread, have you tested this locally yet? with a `~/.mackup/slack.cfg`file or similar

Based on the directory listing I'm guessing the most meaningful things might be...

```
Cookies
Cookies.journal
Preferences
local-settings.json
```

but I couldn't tell you offhand... since it's an electron app afaik I'm assuming the cookies are the logged-in state. 

Would love to see if something like this worked, syncing Slack workspaces between computers sucks right now!

lra:
Dupe of https://github.com/lra/mackup/pull/970

PR #1243, opened 2018-10-29
Add support for Adobe Illustrator CC2019 (v23)

nimbling:
It was already in the readme as a supported application.

• I did some googling but could not find how OSX deals with ".localized" folders.

• I do wonder if the version number in these folders could be a variable?

### DIFF
--- a/mackup/applications/illustrator.cfg
+++ b/mackup/applications/illustrator.cfg
@@ -11,3 +11,21 @@ Library/Preferences/Adobe Illustrator 18 Settings
 Library/Preferences/Adobe Illustrator 19 Settings
 Library/Preferences/Adobe Illustrator 20 Settings
 Library/Preferences/com.adobe.illustrator.plist
+
+Library/Preferences/Adobe Illustrator 23 Settings/en_US/Adobe Illustrator Prefs
+Library/Preferences/Adobe Illustrator 23 Settings/en_US/Workspaces/
+Library/Preferences/Adobe Illustrator 23 Settings/en_US/Modified Workspaces/
+Library/Preferences/Adobe Illustrator 23 Settings/en_US/DataRecovery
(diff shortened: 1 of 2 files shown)


mackup/applications/illustrator.cfg
```diff
@@ -11,3 +11,20 @@ Library/Preferences/Adobe Illustrator 18 Settings
 Library/Preferences/Adobe Illustrator 19 Settings
 Library/Preferences/Adobe Illustrator 20 Settings
 Library/Preferences/com.adobe.illustrator.plist
+
+Library/Preferences/Adobe Illustrator 23 Settings/en_US/Adobe Illustrator Prefs
```
lra: If you look up, the whole folders have been synced. Why not doing the same for 23?

mackup/applications/illustrator.cfg
```diff
@@ -11,3 +11,20 @@ Library/Preferences/Adobe Illustrator 18 Settings
 Library/Preferences/Adobe Illustrator 19 Settings
 Library/Preferences/Adobe Illustrator 20 Settings
 Library/Preferences/com.adobe.illustrator.plist
+
```
lra: Not sure about empty lines in those files, let's just not add them.

mackup/applications/illustrator.cfg
nimbling: Hm! I agree. I wasn't sure if those whole folders *would* get synced. But absolutely, top level folders is what we want.

mackup/applications/illustrator.cfg
nimbling: Agreed!

lra: Can you fix those if you agree? =)

lra:
just need to remove those empty lines

nimbling:
I'd love to comply but don't know how!

lra:
I'll do it, no worries.

PR #1249, opened 2018-10-30
support for calculator program bc

paxperscientiam:


mackup/applications/bc.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = bc
+
+[configuration_files]
+.bcrc
```
lra: Can you end the file with a newline char?

paxperscientiam: should be good now.

PR #1276, opened 2018-12-20
Updated karabiner-elements.cfg

mrymtsk:
Changed the config file so the whole ~/.config/karabiner directory will be backed up.
See: [Manual - Karabiner - Software for macOS](https://pqrs.org/osx/karabiner/document.html#configuration-file-path)


mrymtsk:
@lra Corresponding entry has been added to CHANGELOG.md

PR #1291, opened 2019-01-15
Add sync for Brewfile

caalberts:


caalberts:
Hi, it's actually in https://github.com/Homebrew/homebrew-bundle. Should I rename the config to `homebrew-bundle` to be specific for this?

Sorry for the oversight.

lra:
Yes, it would need this at the minimum. But I'm not sure it's a good fit: per their doc, the `Brewfile` is supposed to be at the root of each project, not at the root of the user's home.

screen shot 2019-01-15 at 8 38 38 pm

So what's your use case?

caalberts:
I'm using it to keep a list of installed brews in my machine. With that, I can easily use `brew bundle install` to install all my brew dependencies across machines.

See:
- https://robots.thoughtbot.com/brewfile-a-gemfile-but-for-homebrew
- https://github.com/christoomey/dotfiles

I agree it may not be the intended use case from the official doc. I'm ok with keeping this out if you think it's not a good fit.

lra:
Yes, sorry but this is more a personal config than files officially supported.
Feel free to save this in your personal config files.

PR #1295, opened 2019-01-19
Add support for Storyist 3

mutantant:
Storyist 3 is a novel writing app. 2 config files, one which contains your license if you purchase, and the other your settings.


mackup/applications/storyist-3.cfg
```diff
@@ -0,0 +1,6 @@
+[application]
+name = storyist-3
+
+[configuration_files]
+Library/Preferences/.com.storyist.storyist.plist
```
lra: This file really exists? Never saw a hidden plist file.

mackup/applications/storyist-3.cfg
```diff
@@ -0,0 +1,6 @@
...
+Library/Preferences/com.storyist.storyist.plist
```
lra: Can you finish the line with a newline character?

mackup/applications/storyist-3.cfg
mutantant: Yes. It's a Storyist specific thing. If purchased, your "key" is in there. Otherwise it's a garbage key which lets Storyist know it's in trial mode. A non-standard way to do things for sure.

mackup/applications/storyist-3.cfg
mutantant: Will do

mackup/applications/storyist-3.cfg
lra: 👌🏼

mackup/applications/storyist-3.cfg
mutantant: Done

PR #1299, opened 2019-01-22
Fixes #1297;remove bundle directory from vim config

### LINKED ISSUE #1297: Why .vim/bundle is back up?
.vim/bundle folder is very big when you install lots of plugs.
why mackup backup this folder? should we just ignore this folder by default ?

cocobear:
We should follow the principle:

> - The configuration should sync the minimal set of data, so that syncing happens quickly. Leave large app data out of the sync configuration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Add support for Storyist 3 writing software (via @mutantant)
 - Add support for WordGrinder (via @mutantant)
 - Fix support for Adobe Illustrator CC2019 (v23)
+- remove `bundle` directory from vim config (via @cocobear)
 
 ## Mackup 0.8.22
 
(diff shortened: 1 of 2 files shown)


malomalo:
Since we are removing `.vim/bundle` should we have another config for a full vim backup?

I mainly use this to just move from and old to a new install and even though the bundle directory might be bigger, it's not huge and it's not happening that often to warrant it be left out. Manually downloading or reinstalling every time was what I was trying to avoid with Mackup

lra:
You can customize the vim config for yourself: https://github.com/lra/mackup/blob/master/doc/README.md#add-support-for-an-application-or-any-file-or-directory

malomalo:
Yea I'm doing that right now, just always have to bring both the `.mackup.cfg` and `.mackup` dir now.

Perhaps it'd be worth having the ability to had the custom configs in the sync dir instead of the user dir so that they follow you to a new computer?

lra:
should be the case if you enable it: https://github.com/lra/mackup/blob/master/mackup/applications/mackup.cfg

PR #1307, opened 2019-01-28
 added support for deepin-terminal 

sUyMur:


sUyMur:
disregard this. Will add the whole deepin folder from .config

Issue #1234, opened 2018-10-16
iTunes Scripts are broken on macOS Mojave


dnicolson:
Since installing macOS Mojave I noticed that if the `~/Library/iTunes/Scripts` directory is a symlink the script menu in iTunes will not appear.

I have made a bug report with Apple, however, all of the AppleScript-related bugs I have submitted over the years have not been fixed or acknowledged.

Perhaps the `iTunesScripts` configuration should be removed for the time being?

lra:
Yeah, if it's not supported anymore, we should remove it. Want to do a PR?

pquerner:
If its only a problem for Mojave, why remove it for all MacOSX versions?

lra:
There is no way to specify a config per MacOS version in Mackup.

PR #1311, opened 2019-01-28
Remove support for iTunes Scripts

dnicolson:
This pull request removes support for iTunes Scripts as they no longer work under macOS Mojave, it resolves https://github.com/lra/mackup/issues/1234.

Any suggestions are welcome, a bug has been opened for half a year and is reflected with this [radar](http://www.openradar.me/radar?id=4944094501535744).


lra:
Looks good, but what's openradar? Can we rely on it?

PR #1316, opened 2019-02-02
add support for rofi

pat-s:


mackup/applications/rofi.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = rofi
+
+[configuration_files]
+.config/rofi/config.rasi
```
lra: You should set your edit to add a newline char at the end of each line, I'll fix this one.

PR #1323, opened 2019-02-17
Update neovim.cfg

zhuxiaoxi:
This directory contains neovim's autoload, which I think should be synchronized.


lra:
Not sure about this, see https://github.com/neovim/neovim/issues/5196#issuecomment-238550438

PR #1325, opened 2019-02-26
add conf for zabbix-cli

dzogrim:
Add conf. backup to Mackup for [zabbix-cli](https://github.com/usit-gd/zabbix-cli) ; tested and works fine


mackup/applications/zabbix-cli.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = Zabbix_CLI
```
lra: `Zabbix-cli`?

PR #1343, opened 2019-03-31
Add support for iStat Menus 6

Dj3ross:


lra:
Thanks, can you merge all istats menus entries?

PR #1346, opened 2019-03-31
Add support for Visual Studio Code OSS and Insiders

horosgrisa:


lra:
I don't understand, where those are coming from? On https://code.visualstudio.com/ there is a way to download the insiders one, but no OSS vs non OSS.

horosgrisa:
OSS build isn't official but some linux distros have only this build, for example https://www.archlinux.org/packages/community/x86_64/code/

PR #1375, opened 2019-05-14
add global git ignore

lorenzofanchi:


lra:
This has been removed before, it's not standard: https://github.com/lra/mackup/pull/351

lorenzofanchi:
so should `.config/git/ignore` be added?

see https://github.com/lra/mackup/pull/926#issuecomment-273964314

lra:
it's already there:
```
[xdg_configuration_files]
git/ignore
```

`xdg_configuration_files` points to your XDG folder, `.config` by default.

lorenzofanchi:
what about when that's not set? running mackup on my machine doesn't back it up..

pvinis:
Well that's kind of a config problem, no? Otherwise let's also put '/Users/lorenzofanchi/.config/git/ignore'

lorenzofanchi:
Not sure.. From the linked post (which draws from https://git-scm.com/docs/gitignore) it reads:

> If $XDG_CONFIG_HOME is either not set or empty, $HOME/.config/git/ignore is used instead.

So I would gather it would actually make sense to include `.config/git/ignore`

No need for witty remarks btw. We're all just here to improve mackup....

lra:
if `XDG_CONFIG_HOME` is not set, Mackup act as if it was set to `~/.config`, the expected default.

> I would argue it would actually make sense to include .config/git/ignore

It's what's already defined in the config file. If it's not working, it's a bug.

lorenzofanchi:
Tested the develop version and everything now works as expected

PR #1382, opened 2019-05-23
Add lftp

dzogrim:
This PR is about add support for lftp.rc ; thanks.

### DIFF
--- a/mackup/applications/lftp.cfg
+++ b/mackup/applications/lftp.cfg
@@ -1,0 +1,5 @@
+[application]
+name = lftp
+
+[configuration_files]
+.config/lftp/rc
--- a/mackup/applications/lftp.cfg
+++ b/mackup/applications/lftp.cfg
@@ -2,4 +2,7 @@
 name = lftp
 
 [configuration_files]
-.config/lftp/rc
+rc
+
+[xdg_configuration_files]
+.config/lftp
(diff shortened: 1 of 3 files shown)


mackup/applications/lftp.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = lftp
+
+[configuration_files]
+.config/lftp/rc
```
lra: You should leverage the XDG config folder, see https://github.com/lra/mackup/blob/master/mackup/applications/tig.cfg#L7-L8 for an example.

dzogrim: OK! thanks @lra I made modifications as suggested…

dzogrim: 👍

mackup/applications/lftp.cfg
```diff
@@ -0,0 +1,8 @@
+[application]
+name = lftp
+
+[configuration_files]
+rc
```
lra: What's this one?

mackup/applications/lftp.cfg
```diff
@@ -0,0 +1,8 @@
...
+
+[xdg_configuration_files]
+.config/lftp
```
lra: You don't have to specify the XDG folder, I suggest trying your conf until it works before opening a PR.

Thanks!

Issue #1386, opened 2019-06-04
Mackup breaks on macOS Catalina DP1 when using iCloud (path change)


fzslm:
Prior to 10.15, the path for iCloud Drive was `~/Library/Mobile Documents/com~apple~CloudDocs`. In 10.15, this path has changed to `~/Library/CloudStorage/iCloud Drive`. Upgrading to Catalina causes old symlinks to break. Some sort of migration process will be necessary for these users in time for the release.

jnbn:
For a temporary solution you can try listing them by
```
s -la ~/ | grep "\->"
```

and change all manually.

jnbn:
Thought a bash would be good  but have no time to finish it. If there's anybody who can do it voluntarily it'd be like this much or less.

```bash
#! /bin/bash

DIR= "~"
OLD_PATTERN="/Users/jnbn/Library/Mobile Documents/com~apple~CloudDocs"
NEW_PATTERN= "/Users/jnbn/Library/CloudStorage/iCloud Drive/"

while read -r line
do
    echo $line
    CUR_LINK_PATH="$(readlink "$line")"
    NEW_LINK_PATH="$CUR_LINK_PATH"  
    NEW_LINK_PATH="${NEW_LINK_PATH/"$OLD_PATTERN"/"$NEW_PATTERN"}"
    rm "$line"
    ln -s "$NEW_LINK_PATH" "$line"
done <<< $(find "$DIR" -type l -maxdepth 1) 
```

jnbn:
Seems like an it needs an update on 

https://github.com/lra/mackup/blob/16cc85a4725c95db6ac6031f54223fe390fe5ee1/mackup/utils.py#L306 

```
if 
defaults read loginwindow SystemVersionStampAsString > 10.15
yosemite_icloud_path = '~/Library/CloudStorage/iCloud Drive/
```

joshmedeski:
The new macOS is called "Catalina". So it sounds like Mackup can be updated to detect Catalina and change the iCloud path dynamically.

fearoffish:
Following on from what @jnbn said, I used the following:

```
#!/bin/bash

set -e

OLD_PATTERN="/Users/$USER/Library/Mobile Documents/com~apple~CloudDocs"
NEW_PATTERN="/Users/$USER/Library/CloudStorage/iCloud Drive"

for line in $(find . -type l -maxdepth 1)
do
	if [[ $(readlink $line) == *"CloudDocs"* ]]; then
[...]

PR #1393, opened 2019-06-20
Fix for iCloud path change in Catalina

rsoper:
Using platform to check for macOS version and setting icloud_home to the correct path accordingly. 
Revolves #1386

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@
 
 # Used by pyenv
 /.python-version
+
+# Used by Visual Studio Code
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -10,3 +10,6 @@
 
 # Used by pyenv
 /.python-version
+
```
pvinis: these lines can be in another pr maybe.

rsoper:
Need to retract this temporarily.

PR #1394, opened 2019-06-29
added config file for pnpm

paxperscientiam:
Hey @lra , i have a question. In addition to the files I've marked in the new config file, the pnpm folks indicate that settings are read from npm config files. I decided not to include those as I'm guessing you can't have two config files pointing to the same program config file. Is that accurate?

This address #1389


lra:
No you can, feel free to.

Issue #1264, opened 2018-11-28
Incompatibility with BBEdit 12.5.1


semicolonsnet:
On opening the newest version of BBEdit (12.5.1), I get the attached error message indicating symlinks to the library folder are no longer supported. I've temporarily put it in my ignore list, but would love to find a way to sync BBEdit again.

![screen shot 2018-11-28 at 10 33 54 am](https://user-images.githubusercontent.com/11462102/49174243-a774de00-f2fa-11e8-9295-2b08d6cfe98c.png)

lra:
Sadly, there are only 2 solutions:
- mackup support copying files + watching file changes
- bbedit support linked files again

PR #1398, opened 2019-07-03
Add BBEdit, Beyond Compare, and Many Tricks Butler

dzg:


lra:
1. bbedit is known to not play well, see https://github.com/lra/mackup/issues/1264 please test your additions before a PR
2. Keep one PR per application so we can easily discuss each

Thanks!

PR #1406, opened 2019-07-20
Added support for dejal time out

rafaelgil83:
Love the app guys. I noticed there is another PR for the same app but that's not the location for the backups, so even though I'm new to Python I'm trying to update it the right way. I did what I could and my local test worked. But if some else could test it...


mackup/applications/timeout.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = TimeOut
+
+[configuration_files]
+Library/Group Containers/6Z7QW53WB6.com.dejal.timeout
```
lra: This is a user specific path, this is not gonna work on other workstations.

rafaelgil83: So, afaik it has two versions. I've always used the one in homebrew "brew cask install time-out", and the data location I don't think is user specific. It's actually documented in the app's website: https://www.dejal.com/faq/timeout/data
But up to you, I can close the PR.

lra: Yes, but as it's containerized, it might overwrite mackup's link everytime it's launched.
Can you try?

rafaelgil83: I have it running in my branch with no issues, it stays as a symlink every time I open the app. That full path I added in the setting file is not a random name, it's more like an ID as documented in the official website. https://www.dejal.com/faq/timeout/data

rafaelgil83: Any update with this PR?

lra: yes, missing newline, I'll fix it

Issue #1410, opened 2019-07-26
0.8.25 crash (Mac homebrew)


shawnyeager:
`Traceback (most recent call last):
  File "/usr/local/bin/mackup", line 11, in <module>
    load_entry_point('mackup==0.8.25', 'console_scripts', 'mackup')()
  File "/usr/local/Cellar/mackup/0.8.25/libexec/lib/python3.7/site-packages/mackup/main.py", line 89, in main
    verbose)
  File "/usr/local/Cellar/mackup/0.8.25/libexec/lib/python3.7/site-packages/mackup/application.py", line 26, in __init__
    assert isinstance(files, set)
AssertionError`

Xdeon:
Same here. Help appreciated.

dmitry-ivanov:
Yeah, I have the same issue too.

mnannola:
Same issue.  Using this with Dropbox if that matters.

`➜  ~ mackup backup
Mackup needs a directory to store your configuration files
Do you want to create it now? </Users/mnannola/Dropbox/Mackup> <Yes|No>Yes
Traceback (most recent call last):
  File "/usr/local/bin/mackup", line 11, in <module>
    load_entry_point('mackup==0.8.25', 'console_scripts', 'mackup')()
  File "/usr/local/Cellar/mackup/0.8.25/libexec/lib/python3.7/site-packages/mackup/main.py", line 89, in main
    verbose)
  File "/usr/local/Cellar/mackup/0.8.25/libexec/lib/python3.7/site-packages/mackup/application.py", line 26, in __init__
    assert isinstance(files, set)
AssertionError`

zhongwangninja:
Same issue when executing `mackup backup` `mackup uninstall --f`
```
❯ mackup backup
Traceback (most recent call last):
  File "/usr/local/bin/mackup", line 11, in <module>
    load_entry_point('mackup==0.8.25', 'console_scripts', 'mackup')()
  File "/usr/local/Cellar/mackup/0.8.25/libexec/lib/python3.7/site-packages/mackup/main.py", line 89, in main
    verbose)
  File "/usr/local/Cellar/mackup/0.8.25/libexec/lib/python3.7/site-packages/mackup/application.py", line 26, in __init__
    assert isinstance(files, set)
AssertionError

❯ mackup uninstall --f
Traceback (most recent call last):
  File "/usr/local/bin/mackup", line 11, in <module>
[...]

Issue #1408, opened 2019-07-22
AssertionError on Ubuntu 18.04.2 LTS, Mackup 0.8.25, Python 3.6.7


alancwoo:
I'm trying to `mackup restore` on a machine running

- Ubuntu 18.04.2 LTS
- Mackup 0.8.25
- Python 3.6.7

It fails immediately with the following:

```
Traceback (most recent call last):
  File "/home/REDACTED/.pyenv/versions/3.6.7/bin/mackup", line 10, in <module>
    sys.exit(main())
  File "/home/REDACTED/.pyenv/versions/3.6.7/lib/python3.6/site-packages/mackup/main.py", line 102, in main
    verbose)
  File "/home/REDACTED/.pyenv/versions/3.6.7/lib/python3.6/site-packages/mackup/application.py", line 26, in __init__
    assert isinstance(files, set)
AssertionError
```

I sync via dropbox, and to debug I made a tar.gz of the original mackup folder and copied/extracted it directly with no luck :( Not sure how to proceed to debug further, I've also tried `mackup restore -v` with no luck.

lubocheng:
me too, it may be a spelling error, try to fix with this:
`sudo vi /usr/local/lib/python2.7/dist-packages/mackup/application.py`
go to line 26, note it or replace it with follow:
`assert isinstance(files, list)`
Then it will be ok

dstroot:
@lubocheng Brilliant!  This worked.  Without this fix Mackup is plain broken on python 2 and 3

ileitch:
Same here on macOS with Mackup 0.8.25

misaka:
Duplicate of https://github.com/lra/mackup/issues/1410

Looks like the issue is caused by the change in this commit: https://github.com/lra/mackup/pull/1402/commits/54e5d98ac84e9db10039cd2b28fe5c02b4460541#diff-cfe235ad3bb75c756c89b6a752e6dd4aR142. It seems that adding `sorted` there turns the set into an array.

```python
(Pdb) self.apps[name]['configuration_files']
{'.mackup', '.mackup.cfg'}
(Pdb) sorted(self.apps[name]['configuration_files'])
['.mackup', '.mackup.cfg']
```

which fails the assertion in the stack trace above:

```python
assert isinstance(files, set)
```

misaka:
PR available #1411

PR #1411, opened 2019-07-28
convert sorted appsdb files list to set

misaka:
Fixes breakage in backup & restore commands.

See issues #1410 & #1408

### DIFF
--- a/mackup/appsdb.py
+++ b/mackup/appsdb.py
@@ -139,7 +139,7 @@ class ApplicationsDatabase(object):
         Returns:
             set of str.
         """
-        return sorted(self.apps[name]['configuration_files'])
+        return set(sorted(self.apps[name]['configuration_files']))
 
     def get_app_names(self):
         """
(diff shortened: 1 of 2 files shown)


mackup/appsdb.py
```diff
@@ -139,7 +139,7 @@ def get_files(self, name):
         Returns:
             set of str.
         """
-        return sorted(self.apps[name]['configuration_files'])
+        return set(sorted(self.apps[name]['configuration_files']))
```
lra: The problem is that a set is not sorted by definition, so we'd rather remove the `sorted` until we use sorted set =/

misaka: Make sense. 👍

lra:
Thanks for the PR btw, I just did a fix on my side as it's kind of urgent and I did not know if you'd be online. Sorry for takeover!

misaka:
np, just glad to see it fixed. Thanks for the quick response!

robinvalk:
Any eta on a new homebrew version?

lra:
whenever someone does a PR.

robinvalk:
Ah now I got an update! Yesterday there was no new version available via homebrew. Thanks!!

PR #1424, opened 2019-08-12
Try pipenv

lra:


jpmat296:
https://github.com/jpmat296/mackup/tree/master/doc#locally-test-an-application-before-submitting-a-pull-request

I think `pipenv` must be used after `make develop`. I can submit PR if you confirm. Thanks

PR #1427, opened 2019-08-15
Add Bat Application

joshmedeski:


mackup/applications/bat.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = Bat
+
+[configuration_files]
+.config/bat/config
```
lra: `.config` is just a default for XDG config files. Can you use the `xdg_configuration_files` section instead? See `https://github.com/lra/mackup/blob/master/mackup/applications/git.cfg` for an example.

PR #1434, opened 2019-08-24
feature: add support for proselint

paxperscientiam:


mackup/applications/proselint.cfg
```diff
@@ -0,0 +1,6 @@
+[application]
+name = proselint
+
+[configuration_files]
+.proselintrc
+config/proselint/config
```
lra: This is missing a staring dot I think
And this should live in the `xdg_configuration_files` section as `.config` is just a default that can be changed, see https://github.com/lra/mackup/blob/master/mackup/applications/tig.cfg for an example.

There is no doc about it yet sorry.

paxperscientiam: WRT `xdg_configuration_files`, is the idea to improve linux compatibility?

Issue #1436, opened 2019-08-24
Feature request: "n or less" wildcard support


paxperscientiam:
Some applications namespace configuration paths by version.

For example, Gimp does this:
Gimp 2.8 -- "~/Library/Application Support/GIMP/2.8"
Gimp 2.10 -- "~/Library/Application Support/GIMP/2.10"

etc

lra:
Yes, Mackup only support contant paths right now, but we could imagine a parsing engine with simple variables in the paths like support for 1 `?` or multiple `*` characters.

pshirshov:
Wildcards would be very useful, could you add them?

semkagtn:
Plese, check: https://github.com/lra/mackup/pull/1581

PR #1437, opened 2019-08-24
feature: support for gimp

paxperscientiam:
This is a replacement for PR #1172 

Please see issue #1436 re future improvements for gimp support and for applications with similar directory structures

@fharper this may interest you.


paxperscientiam:
@lra, just being cautious. I am not 100% sure that it covers all resource files. I tried asking around and digging around, but I was unable to find a definitive list.

If I find anything missing going forward, I'll try to add it.

Thanks.

Issue #1446, opened 2019-09-02
Robomongo vs Robo 3T


captnswing:
The Robomongo app has been renamed to Robo 3T. The config files are now in a different place, as well. Should I

1) rename `robomongo.cfg` -> `robo3t.cfg` and change fields accordingly
2) add a new `robo3t.cfg` 

?

lra:
I'd go with 1.

PR #1448, opened 2019-09-02
Add support for Robo 3T

captnswing:
as discussed in https://github.com/lra/mackup/issues/1446

### DIFF
--- a/mackup/applications/robomongo.cfg
+++ b/mackup/applications/robomongo.cfg
@@ -1,5 +1,0 @@
-[application]
-name = Robomongo
-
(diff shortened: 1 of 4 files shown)


lra:
This was good. Can you reopen it?

mackup/applications/robomongo.cfg
```diff
@@ -1,5 +0,0 @@
-[application]
-name = Robomongo
-
-[configuration_files]
-.config/robomongo
```
lra: I would keep this and copy it in the renamed file.

Issue #1447, opened 2019-09-02
Auto-generate list of supported Applications


captnswing:
by adding a `homepage` field to the `[application]` stanza, a markdown file (e.g, `SUPPORTED_APPS.md` could be auto-generated during the build process. This could be used to off-load information from the README and keep information in markdown file and the available config files in the applications directory in sync

PR #1449, opened 2019-09-02
Auto-generate list of supported Applications

captnswing:
@ira first stab at the idea expressed in https://github.com/lra/mackup/issues/1447

I'd say if you decided to go ahead with this, what would be left TODO in this PR would be

- [ ] decide final name and location of `SUPPORTED_APPLICATIONS.md`
- [ ] find ways to run `generate_supported_apps.py` during CI?
- [ ] and/or update `CONTRIBUTING.md` with additional instructions

Let me know what your thoughts are. This is just a suggestion!


lra:
I like this, we can start with `APPLICATIONS.md`.
For running the script in CI, this is easy, look at how we run other stuff, e.g. https://github.com/lra/mackup/blob/master/.circleci/config.yml#L84

But you don't want to generate the file on CI, you want the script to have a "diff" mode that's gonna return non-zero (and fail CI) if the list is out of date. And you can add a Makefile entry to generate the up2date `APPLICATIONS.md` file in a PR.

Feel free to continue working on this PR.

PR #1453, opened 2019-09-06
feat: add support for Starship

foray1010:


mackup/applications/starship.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = Starship
+
+[configuration_files]
+.config/starship.toml
```
lumaxis, line 5: I believe this should use the [built-in XDG support](https://github.com/lra/mackup/tree/master/doc#add-support-for-an-application-using-the-xdg-directory) instead: 
```suggestion
[xdg_configuration_files]
starship.toml
```

foray1010: seems starship is not using XDG: https://github.com/starship/starship/issues/896

foray1010:
already supported in https://github.com/lra/mackup/blob/master/mackup/applications/starship.cfg, closing

mackup/applications/starship.cfg
lumaxis: Oh, indeed. I just assumed it was since the file is in `.config`. Sorry about that!

lra: Yeah it's super confusing.

PR #1456, opened 2019-09-08
Add support for Music

dnicolson:
This pull request adds support for scripts in the macOS 10.15 Music app.


mackup/applications/music.cfg
```diff
@@ -0,0 +1,7 @@
+[application]
+name = Music
+
+[configuration_files]
+Library/Music/Scripts
```
lra: Apple software have an history of not behaving well with Mackup. What's in this folder?

dnicolson: OSA scripts just like iTunes. It is true that from 10.4.0 – 10.4.3 the symlink caused iTunes scripts to [not work](https://github.com/lra/mackup/pull/1311), it has since been [resolved](https://github.com/lra/mackup/pull/1349) though.

lra: 👌

lra:
They call it Apple Music, changed your PR a little.
Screen Shot 2020-02-27 at 1 53 42 PM

PR #1459, opened 2019-09-11
Add .default-npm-packages to asdf.cfg

caalberts:
Default set of npm packages to be installed with NodeJS.
https://github.com/asdf-vm/asdf-nodejs#default-npm-packages


lra (approved):
Thanks

lra:
Looks like this was already added, but merging anyway for the order.

PR #1461, opened 2019-09-13
Add support for rclone

dnicolson:
This pull request adds support for [rclone](https://rclone.org) configuration.


mackup/applications/rclone.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = rclone
+
+[configuration_files]
+.config/rclone
```
lra: This needs to be moved under `xdg_configuration_files`, see https://github.com/lra/mackup/tree/master/doc#add-support-for-an-application-using-the-xdg-directory

I'll do it.

lra:
Thanks

PR #1463, opened 2019-09-17
Add ability to backup "defaults" settings

eelmasllari:
Some applications, e.g. OTP Manager, Vanilla, etc. do not have config files but save their configuration using the Mac OS "defaults" system, e.g. "defaults write nl.stickybits.OTP-Manager <some-key> <some-value>"
This PR adds the ability to backup and restore the settings of such applications in the same way as backing up and restoring configuration files.

### DIFF
--- a/doc/.mackup.cfg
+++ b/doc/.mackup.cfg
@@ -41,6 +41,12 @@ engine = dropbox
 # engine you chose, e.g. "~/Dropbox/Mackup".
 directory = Mackup
 
+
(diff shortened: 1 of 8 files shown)


eelmasllari:
The same function can be achieved by saving the plist file from under ~/Library/Containers/<your- app>/Data/Library/Preferences/<your-app>.plist. This PR therefore adds no new functionality and can be ignored.

PR #1464, opened 2019-09-18
Add systemd, nrm and fix bat

grigorii-horos:


lra:
One application per PR please.

PR #1479, opened 2019-10-27
Add support for Vivaldi Browser

ttuygun:
Caches of Vivaldi located in Library/Caches/Vivaldi
It sync all of session data with extension settings and some of the settings that are not synced via Vivaldi sync

### DIFF
--- a/README.md
+++ b/README.md
@@ -521,6 +521,7 @@ See the [README](doc/README.md) file in the doc directory for more info.
 - [Ventrilo](http://www.ventrilo.com/)
 - [Versions](http://www.versionsapp.com)
 - [Vim](http://www.vim.org/)
+- [Vivaldi](https://vivaldi.com/)
 - [Vimperator](http://www.vimperator.org/vimperator)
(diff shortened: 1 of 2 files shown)


ttuygun:
I deeply investigate that all of the user settings stored at Default folder in Application Support/Vivaldi. This default folder only sync that currently active user of Vivaldi. If user create more than one user it will be created new folders like Profile 1, Profile 2, etc. At the point, if user delete a profile, the system already continues the latest created profile like Profile 3. For this reason, no need to sync another folder except Default one. Why sync user settings if Vivaldi sync available? Because Vivaldi sync only small parts of settings and some extension installations. But this folder, sync everything including extension data, all of the user settings, mouse gestures, search options. @lra

lra:
Isn't it a bug in the Vivladi browser if it has a partially working sync system?
Also, can you add a line to the changelog, like stated in https://github.com/lra/mackup/blob/master/.github/CONTRIBUTING.md

Thanks.

lra:
Closing, this is handled by https://help.vivaldi.com/desktop/tools/user-profiles/#Sync_profiles

PR #1480, opened 2019-10-28
TripMode support

ttuygun:
It sync paddle file for license and application settings such as custom profile schedular times which saved in sqlite file

image


mackup/applications/tripmode.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = TripMode
+
+[configuration_files]
+Library/Application Support/TripMode
```
lra: missing newline

lra: https://github.com/lra/mackup/pull/1480/commits/b1a7bdc278fc4bf3f272706e874390f2ed161167

PR #1482, opened 2019-10-29
Add Starship support

callummr:
🚀


callummr:
The lint failures are on files I didn't touch..?

mackup/applications/starship.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = Starship
+
+[configuration_files]
```
lra: This should be moved under `xdg_configuration_files`, see https://github.com/lra/mackup/tree/master/doc#add-support-for-an-application-using-the-xdg-directory

callummr: Does this still apply when the application consuming the config file doesn't look for `XDG_CONFIG_HOME`, but hardcodes `.config` as the path it reads from? That's what Starship's doing here: https://github.com/starship/starship/blob/1eabd527252291398df5749d30f3459ea2a03823/src/config.rs#L184

lra: heh, no it doesn't.
Weird.

hrvstr:
Am I doing something wrong here? I am using Mackup 0.8.28 (installed with home-brew) and I don't see starship when I do `mackup list`. I tried adding it to my config anyway and got this:

```
  File "/usr/local/bin/mackup", line 11, in <module>
    load_entry_point('mackup==0.8.28', 'console_scripts', 'mackup')()
  File "/usr/local/Cellar/mackup/0.8.28_1/libexec/lib/python3.8/site-packages/mackup/main.py", line 86, in main
    app = ApplicationProfile(mckp, app_db.get_files(app_name), dry_run, verbose)
  File "/usr/local/Cellar/mackup/0.8.28_1/libexec/lib/python3.8/site-packages/mackup/appsdb.py", line 141, in get_files
    return self.apps[name]["configuration_files"]
KeyError: 'starship'
```

lra:
It's not in a released version yet, I plan to release one tonight.

PR #1483, opened 2019-10-31
Update Go2Shell bundle identifier

dnicolson:
This pull request updates the bundle identifier for Go2Shell.

The latest version in the Mac App Store is 1.2.2 and uses the old bundle identifier. It is not compatible with Catalina but can be installed with `mas install 445770608`.

The latest release is 2.5 and uses the new bundle identifier, it can be installed with `brew cask install go2shell`.


dnicolson:
The version history is not readily available, but I would say it's safe to ignore the old bundle identifier as it is from at least 5 years ago and no longer compatible. Also, it seems that there is only one key in the plist file that is important anyway.

lra:
Thanks

PR #1492, opened 2019-11-28
added support for mitmproxy

etiennelb:


mackup/applications/mitmproxy.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = mitmproxy
+
+[configuration_files]
+.mitmproxy/config.yaml
```
lra: newline missing

lra: https://github.com/lra/mackup/pull/1492/commits/e5098187a9ed2e1831af48c66519fdfe86af3977

PR #1494, opened 2019-11-29
Update bash.cfg to add .bash_history

AdrianSimionov:


lra:
Not sure if it makes sense to sync `.bash_history`. It's host specific data, and not config.

PR #1503, opened 2019-12-20
Add Poetry Support

ryoppippi:


mackup/applications/poetry.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = poetry
+
+[configuration_files]
+Library/Application Support/pypoetry/config.cfg
```
lra: I can't find any reference to this file in the source code, how did you find this?

ryoppippi: https://github.com/python-poetry/poetry/blob/master/docs/docs/configuration.md
It's mentioned here

lra: 👌

PR #1506, opened 2020-01-13
Add config.yml for cheat

hezhizhen:


mackup/applications/cheat.cfg
```diff
@@ -2,4 +2,5 @@
 name = Cheat
 
 [configuration_files]
+.config/cheat/conf.yml
```
lra: This should be in `xdg_configuration_files`, I'll fix it. https://github.com/lra/mackup/tree/master/doc#add-support-for-an-application-using-the-xdg-directory

PR #1510, opened 2020-01-15
Add VSCode Extension support

ionlights:
Is there a particular reason that `mackup` doesn't backup VSCode extensions since that's needed to transport a total VSCode configuration?

https://code.visualstudio.com/docs/editor/extension-gallery#_common-questions

Extensions on MacOS/Linux are stored in `.vscode/extensions`.

### DIFF
--- a/mackup/applications/vscode.cfg
+++ b/mackup/applications/vscode.cfg
@@ -10,3 +10,5 @@ Library/Application Support/Code/User/settings.json
 .config/Code/User/snippets
 .config/Code/User/keybindings.json
 .config/Code/User/settings.json
+
+.vscode/extensions


rasmusbe:
I think you can include the whole `.vscode` directory, that includes both extensions and permanent command line arguments

ionlights:
Hmm, that's fair. Actually something I've noticed following some further testing and transfers to/from Linux.

Although, upon resetting my mac, I found a needed to symlink some of the files in `Library/Application Support/Code` to get things working properly. I'm not sure of VSCode's plans to actually migrate everything to `.vscode` or support `$XDG_CONFIG_HOME`.

rasmusbe:
We're probably best of by including both `.vscode` and `Library/Application Support/Code`

foray1010:
did you check https://github.com/lra/mackup/pull/1170?

ionlights:
@foray1010 i didn't see that PR, no. though i get what they're saying.

lra:
It was removed in https://github.com/lra/mackup/pull/1170
If you really want to sync those, can you use a user config? See https://github.com/lra/mackup/tree/master/doc#add-support-for-an-application-or-any-file-or-directory

PR #1511, opened 2020-01-16
Added support for AndroidStudio3.3 & Pet

r3a1d3a1:
I'm not sure how things were back when `Library/Application Support/AndroidStudio1.3` was considered for "Android Studio 1.3". But `Library/Application Support/AndroidStudio3.3` takes close to 100MBs in my case and none of the files looked worthy of backup. They were mostly downloaded content.

### DIFF
--- a/README.md
+++ b/README.md
@@ -437,6 +437,7 @@ See the [README](doc/README.md) file in the doc directory for more info.
 - [Pear](http://pear.php.net/)
 - [Pentadactyl](http://5digits.org/pentadactyl/)
 - [Perl](https://www.perl.org/)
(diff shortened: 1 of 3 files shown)


lra:
Can you split both application in a dedicated PR each?

r3a1d3a1:
Hi Laurent. Thank you for your consideration. Unfortunately, I don't have time for that.

lra:
ok, feel free to open another PR when you have time.

r3a1d3a1:
I never will. Especially on this case :)
If you really cared, you would take care of it here, and I certainly don't care about it more than you do.

PR #1512, opened 2020-01-17
Linux support for PhpStorm config

mol91:


lra:
Wow... thanks I guess 😅
What's in those folders?

mol91:
Configuration files for PhpStorm on Linux :smile:

lra:
Thanks

PR #1517, opened 2020-01-21
Improve support for AWS CLI, only sync config and exclude credentials

halilim:
AWS CLI [stores](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-configure-files-where) config in `~/.aws/config` and credentials in `~/.aws/credentials`.

Workaround until this is merged:

```ini
; ~/.mackup.cfg
[applications_to_ignore]
aws
```

```ini
; ~/.mackup/my-files.cfg
[configuration_files]
.aws/config
```

```bash
rm ~/.aws
mv  ~/Library/Mobile\ Documents/com\~apple\~CloudDocs/Mackup/.aws ~
mackup backup
```

And then please consider rotating AWS keys that are stored in `~/.aws/credentials`, depending on your Mackup storage and security requirements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## WIP
 
+- Improve support for AWS CLI, only sync config and exclude credentials (via @halilim)
 - Added support for Bat (via @joshmedeski)
 
 ## Mackup 0.8.27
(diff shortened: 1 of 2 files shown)


lra:
That's a behavior that's expected by users.
You can override the `aws` backup behavior by putting this config file in your `~/.mackup` folder.

quantonganh:
> That's a behavior that's expected by users.

Same as SSH private key, I don't think we should sync aws credentials by default.

PR #1521, opened 2020-01-25
add support for zsh aliases file

udaraau:
adding support for .zsh_aliases file which is commonly used for defining additional aliases along with zsh.


lra:
If it's officially supported by ZSH, it should be added to the `zsh` config.
If it's not, you should keep it as a user config.

PR #1522, opened 2020-02-02
update support for RStudio to handle JSON conf file

petrbouchal:
The latest version of RStudio [stores its conf in a JSON](https://docs.rstudio.com/ide/desktop-pro/latest/settings.html).

This PR addresses this change and also includes other config files in the backup.

### DIFF
--- a/mackup/applications/rstudio.cfg
+++ b/mackup/applications/rstudio.cfg
@@ -4,3 +4,4 @@ name = RStudio
 [configuration_files]
 .rstudio-desktop
 Library/Preferences/org.rstudio.RStudio.plist
+.config/rstudio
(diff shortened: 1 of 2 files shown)


mackup/applications/rstudio.cfg
```diff
@@ -4,3 +4,4 @@ name = RStudio
 [configuration_files]
 .rstudio-desktop
 Library/Preferences/org.rstudio.RStudio.plist
+.config/rstudio
```
lra: To support Freedesktop standard, It should be
```
[xdg_configuration_files]
rstudio
```

e.g. https://github.com/lra/mackup/blob/755095e590deab8f4a2875e36800a021094f683d/mackup/applications/git.cfg

petrbouchal: Thanks - was not aware of this. Updated in 1ce5dde

PR #1240, opened 2018-10-27
Add support for Calibre

ttuygun:

PR #1527, opened 2020-02-25
Add support for Calibre

bodak:
> Adding MacOS and unix support for Calibre

Just a quick PR to add in Calibre sync since #1240 seems to have gone stale.


mackup/applications/calibre.cfg
```diff
@@ -0,0 +1,6 @@
+[application]
+name = Calibre
+
+[configuration_files]
+Library/Preferences/calibre
+.config/calibre
```
lra: Can you use the `xdg_configuration_files` for those? See https://github.com/lra/mackup/blob/master/doc/README.md#add-support-for-an-application-using-the-xdg-directory

PR #1536, opened 2020-02-29
Added support for SBT 1.0

austek:
Added support for SBT 1.0


lra:
thanks, I'll add an entry to the changelog

PR #1538, opened 2020-03-04
sync .gitignore and .gitignore_global

ryoppippi:


jamiew:
I haven't seen a `~/.gitignore_global` before, is that just an alternative configuration or a custom setup? `git config --global core.excludesfile` by default is `~/.gitignore`

I might recommend keeping that in a custom ~/.mackup/ config if it is a custom thing

ryoppippi:
I see
Okay I’ll close this

jamiew:
The `~/.gitignore` seems useful, I was just asking about `~/.gitignore_global`! Sorry if I was terse

It looks like we used to support `.gitignore` but no longer? https://github.com/lra/mackup/blob/master/mackup/applications/git.cfg

lra:
It's been brought back multiple times: https://github.com/lra/mackup/pull/926#issuecomment-273964314

PR #1543, opened 2020-03-14
Added support for Joplin

geekrainy:


mackup/applications/joplin.cfg
```diff
@@ -0,0 +1,6 @@
+[application]
+name = Joplin
+
+[configuration_files]
+.config/joplin-desktop/userchrome.css
```
lra: Can you replace those with the XDG directory? See https://github.com/lra/mackup/blob/master/doc/README.md#add-support-for-an-application-using-the-xdg-directory

PR #1548, opened 2020-03-30
Add unified iStat Menus config

lumaxis:
Closes #1343

Addressing [@lra's comment](https://github.com/lra/mackup/pull/1343#issuecomment-478653992) from the above mentioned PR, this adds a unified config for iStat Menus across multiple versions.

I simply removed `istat-menus-5` since I'm not sure what the best way to "merge" configs is 🤔Happy to revert that commit though since this would be a breaking change for people and I'm not sure that is desired.


lumaxis:
@lra Did you get a chance to look at this yet? Any thoughts?

PR #1566, opened 2020-04-23
Add support for Sway (wm)

mk-atlassian:
Includes the whole sway config directory since there are often supporting scripts or config files.


mk-atlassian:
Please give thoughts on which name should be used where. It tends to go by "Sway", "swaywm", and "sway". My impression is the name is "Sway", but its name to avoid collisions is "swaywm".

lra:
Sway is fine, they seem to use it over swaywm.

mk-atlassian:
> Can you add a line to the changelog?

done

PR #1567, opened 2020-04-23
Add support for waybar

mk-atlassian:
Includes the whole waybar config directory so that common supporting scripts are included.


mk-atlassian:
Is there a way to ignore a file in the config? A common setup would include a `temp` file which is just a symlink to a CPU core temp file so that waybar can display the temperature. This is not a big deal to sync since it's just a file with a number. I also do not mind adding this ignore functionality to mackup

lra:
> Is there a way to ignore a file in the config?

No, it's not possible because of the way linking directories work. You can't not have a file while the parent directory is linked.

This might be possible only once the copy functionality (vs link) is implemented.

mk-atlassian:
> No, it's not possible because of the way linking directories work.
Of course, apologies, I wasn't thinking.

My opinion is:
There is a less elegant way which is to symlink all the children of a directory except for the ignored files on each `mackup backup`. This isn't elegant, and it could be unexpected by the user or even contributors adding support for an application.

Copying is nice because the assumption is that the files copied will be pretty small. If there is some trigger though, for copying, there may as well be a trigger for symlinking children?

This discussion should probably be on an issue, is there a one that you'd like to have it on?

lra:
> There is a less elegant way which is to symlink all the children of a directory except for the ignored files on each mackup backup.

This can't work as there are subfolders sometimes.

PR #1568, opened 2020-04-23
Add support for mako

mk-atlassian:
Only includes the config file since it's unlikely that there will be other supporting config files.


mk-atlassian:
> Can you add a line to the changelog?

done

mackup/applications/mako.cfg
```diff
@@ -0,0 +1,6 @@
+[application]
+name = mako
+
+[xdg_configuration_files]
+mako/config
+
```
lra, line 6: extra newline not needed.

PR #1569, opened 2020-04-23
Add support for termite

mk-atlassian:
Only include the config file and not the whole directory since it's unlikely there is supporting files.


mk-atlassian:
Same question as in around single config file support #1568

mk-atlassian:
> Thanks, can you add a line to the changelog?

Hey, is there anything else? All the other PRs were also updated

PR #1572, opened 2020-05-06
added config for SpaceVim

shivams:
SpaceVim is a Spacemacs inspired vim distribution. It stores all of its config inside `~/.SpaceVim.d` folder. The distribution itself resides in `~/.SpaceVim` and `~/.vim` is symlinked to `~/.SpaceVim`.


lra:
implemented in #1602

PR #1580, opened 2020-05-26
ableton-live

semkagtn:


mackup/applications/ableton-live.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = Ableton Live
+
+[configuration_files]
+Library/Preferences/Ableton
```
semkagtn, line 5: Oh I just noticed that this directory contains more then just a config. And this dir is a quite big.

The problem is this app are using versioning like this:
`Library/Preferences/Ableton/Live 10.1.5/Preferences.cfg`.

Wild cards aren't supported, so I close this pull request. In addition it is not a very popular application among programmers.

PR #1590, opened 2020-06-06
GnuPG doesn't sync private keys anymore (#1021)

Bouke:


Bouke:
@lra No problem! There's also this: https://github.com/lra/mackup/blob/master/doc/.mackup/gnupg.cfg, but I'm not sure what to do with it.

lra:
What's the issue?

Bouke:
It is still referencing the private keys.

PR #1592, opened 2020-06-09
remove fish_variables

ho-oto:
fish_variables is automatically generated by fish-shell and is not intended to be edited by a human. Thus, it should not be backed up.

https://fishshell.com/docs/current/index.html?highlight=fish_variable#variables-universal

### DIFF
--- a/mackup/applications/fish.cfg
+++ b/mackup/applications/fish.cfg
@@ -4,6 +4,5 @@ name = Fish
 [xdg_configuration_files]
 fish/config.fish
 fish/conf.d
-fish/fish_variables
 fish/functions
 fish/completions


lra:
Then what's the way to backup a modified fish_user_paths in fish? You can't save `fish_variables` but you can't put the addition `fish_user_paths` to `config.fish` either.

What's the clean way? Append to the PATH variable in `config.fish`?

ho-oto:
Now I use `config.fish` file to append new variables to `PATH`.
If `set -U fish_user_paths foo` or `set -x PATH foo $PATH` is written in `config.fish`, `foo` will be appended to the top of `PATH`.

But, I think it is not a standard way. So, I will make local `fish.cfg` file and close this PR. Sorry for bothering you.

lra:
You are not bothering me, it's a legit question, and their documentation is not clear.
It is so unclear that a long thread about this has lived over years and had been closed: https://github.com/fish-shell/fish-shell/issues/527

I'm open to ideas.

It seems the consensus is to add this to `config.sh`: `contains $fish_user_paths /path/to/add; or set -Ua fish_user_paths /path/to/add`?

jimblue:
My 2cents...

Because  **mackup** backup `fish_variables`, I have to handle two different branch on my dotfiles repo (Linux an Mac) and this is really a pain. 😭

It's also worth mentioning that it's possible to add `fish_add_path /path/to/add` in `config.fish` to extend default `PATH` and save the result to `fish_variables`. Because this is done in `config.fish`, it can be done dynamically and it's possible to adapt the path in `fish_variables` to different OS.

That's why in my opinion, `fish_variables` should be backup.

PR #1597, opened 2020-06-18
docker config migrated

dhb52:
`~/.dockercfg` migrated to `~/.docker/config.json`

docker uses `~/.docker/config.json`
dockerd uses `~/.docker/daemon.json`


dhb52:
IMHO, move `application` folder into submodule is a good way to seperate  `main` and `config`

mackup/applications/docker.cfg
```diff
@@ -2,5 +2,5 @@
 name = Docker
 
 [configuration_files]
-.dockercfg
 .docker/config.json
+.docker/daemon.json
```
lra, line 6: I can't find any ref to this file, and the doc suggests to not touch any other file than the config.json.
https://docs.docker.com/engine/reference/commandline/cli/#configuration-files

What's the rationale?

dhb52: 1. docker (cli) uses ~/.docker/config.json
By default, the Docker command line stores its configuration files in a directory called .docker within your $HOME directory. [ref](https://docs.docker.com/engine/reference/commandline/cli/#configuration-files)

2. dockerd uses ~/.docker/daemon.json
this one I can't find support doc. But when I change config from macOS Docker Desktop UI, it reflects to the file content in `~/.docker/daemon.json`. 
While the official doc more likely describe the rationale on Linux native docker config.

lra: ok that's fine, we can remove it if it breaks.

PR #1602, opened 2020-07-02
Adding SpaceVim to Applications

ionlights:
[SpaceVim][sv] is a `vim` distribution that tries to mimic Spacemacs, but for `vim`. Configurations are stored in either:
1. `~/.SpaceVim.d/`
1. `$SPACEVIMDIR` ([see ref][sv-config])

The default is `~/.SpaceVim.d/`, so that's what makes sense for Mackup to support.

[sv]: https://spacevim.org
[sv-config]: https://spacevim.org/documentation/#custom-configuration


lra:
Can you follow https://github.com/lra/mackup/blob/master/.github/CONTRIBUTING.md ?
Thanks

ionlights:
Migrated computers which led to some issues with upstreaming just from my original branch – I can revert if necessary.

Otherwise, the changes from CONTRIBUTING should be there. 🙂

CHANGELOG.md
```diff
@@ -2,10 +2,12 @@
 
 ## WIP
 
+- Added support for SpaceVim (via @ionlights)
 - Added support for clashX (via @awkj)
 - Added support for Brave (via @cbenv)
 - Added support for Nushell (via @leesiongchan)
 - Updated support for Clipy (via @jclerc)
+- Added support for aerc (via @Crocmagnon)
```
lra, line 10: I don't think this is supposed to be part of this PR

ionlights:
Sorry, needed to revert and overwrite changes. Force-pushing to my own repo seemed to be the simplest. Unsure if there's "a better way" – if so, could you point me in the right direction and I'll do that? 🙂

PR #1603, opened 2020-07-04
Add support for aerc

Crocmagnon:
Supports syncing config files for aerc email client


mackup/applications/aerc.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = aerc
+
+[configuration_files]
+Library/Preferences/aerc
```
lra, line 5: missing newline, I'll fix it

PR #1605, opened 2020-07-23
add support for espanso

maxandersen:
espanso store its user config in Library/Preferences/espanso where there is a `config.yml` file and `user` folder.
All documented at https://espanso.org/docs/configuration/


mackup/applications/espanso.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = espanso
+
+[configuration_files]
+Library/Preferences/espanso
```
lra, line 5: missing newline, I'll add it

PR #1632, opened 2020-09-10
Fix poetry config backup

ameyuuno:
Configuration of poetry is stored in `.toml` file, not `.cfg`.

Based on its documentation https://python-poetry.org/docs/configuration/

> Poetry can be configured via the config command (see more about its usage here) or directly in the config.toml file


lra (approved):
Thanks

lra:
I'll add the unix path.

PR #1655, opened 2020-11-16
Add themes and bundles for Sequel Pro

otherguy:
Custom themes and especially bundles for [Sequel Pro](https://www.sequelpro.com) should be synced as well.


otherguy:
This PR is still up to date :)

lra (approved):
Thanks!

PR #1658, opened 2020-11-17
feat: new app npmrc

jdvivar:
Adding support for npm package npmrc (https://github.com/deoxxa/npmrc), a utility to manage different `.npmrc` files for different uses.

Bear in mind `.npmrcs` is a folder with files, each representing a profile. I have tested and it works!


lra (approved):
Thanks, will add the changelog entry

PR #1671, opened 2021-01-01
fix: aria2 supports XDG_CONFIG_HOME

hongqn:
Quote from [aria2 documentation](https://aria2.github.io/manual/en/html/aria2c.html#aria2-conf):

> By default, aria2 checks whether the legacy path
> `$HOME/.aria2/aria2.conf` is present, otherwise it parses
> `$XDG_CONFIG_HOME/aria2/aria2.conf` as its configuration file.


lra (approved):
thanks

Issue #1621, opened 2020-08-12
Ask cloud storage provider when running Mackup for the first time


nuno-agostinho:
When running Mackup for the first time (e.g. if there is no `.mackup.cfg`), ask the user which cloud provider to use.

Every time I install Mackup in a new computer, I have to check the GitHub to see how to change it to use iCloud. Mackup could simply ask the user which storage provider s/he wants to use and create a `.mackup.cfg` accordingly.

Also, the error message `Error: Unable to find your Dropbox install =(` makes it sound like Mackup only supports Dropbox.

PR #1682, opened 2021-01-08
Improve the error message when no storage is found

Crocmagnon:
This relates to issue #1621. It improves the error message when the storage is not found by pointing the user to the docs.

This is useful for first-time users as they could otherwise just understand that only Dropbox is supported.

I don't know if it closes the issue but at least it improves the situation described.


lra (approved):
Makes sense

lra:
Can you check the CI errors?

Crocmagnon:
Yup, I'm on it.

Crocmagnon:
Done, I cleaned up the commits for the sake of your history.

PR #1691, opened 2021-02-03
Remove-discontinued-service

AugustasV:
Copy was discontinued on May 1, 2016. So I had to remove it from Supported storages list.


joshmedeski:
We'll want to make sure we go through the codebase and remove any copy-related implementation too.

lra:
Thanks!

Issue #1317, opened 2019-02-03
Follow the XDG spec for configuration


pcr910303:
Currently the location of configuration file for mackup is `~/.mackup.cfg`, which clutters the `$HOME` directory. I believe that mackup should closely follow the [XDG specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) ([Summary](https://wiki.archlinux.org/index.php/XDG_Base_Directory)) and allow users to place configuration files to `$XDG_CONFIG_HOME/mackup/config`.

lra:
This is a good idea.

treee111:
Ist done, PR's merged and can be closed?

NorthIsUp:
#1987 to have mackup store its own files in XDG compliance

PR #1694, opened 2021-02-08
Update the config dir path

GentleHoneyLover:
To comply with XDG Base Directory specification: https://wiki.archlinux.org/index.php/XDG_Base_Directory

### DIFF
--- a/mackup/constants.py
+++ b/mackup/constants.py
@@ -16,10 +16,10 @@ MACKUP_APP_NAME = "mackup"
 MACKUP_BACKUP_PATH = "Mackup"
 
 # Mackup config file
-MACKUP_CONFIG_FILE = ".mackup.cfg"
+MACKUP_CONFIG_FILE = ".config/mackup/mackup.cfg"
 
 # Directory that can contains user defined app configs
-CUSTOM_APPS_DIR = ".mackup"
+CUSTOM_APPS_DIR = ".config/mackup"
 
 # Supported engines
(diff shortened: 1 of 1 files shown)


joshmedeski:
I support this change, however, there would have to be a clear way to migrate existing setups (my `.mackup` folder and config file are in the home directory right now).

I believe this would create a problem for people who already have mackup running on their computer.

GentleHoneyLover:
> I support this change, however, there would have to be a clear way to migrate existing setups (my `.mackup` folder and config file are in the home directory right now).
> 
> I believe this would create a problem for people who already have mackup running on their computer.

I'm not a coder so cannot offer a ready solution, but could it be a little migration script as a part mackup that silently moves the config folder and config file to the new location at first run?

Or alternatively, maybe add two locations for config files (legacy location and new location) so that mockup looks for them in both places?

If I knew python I bit better, I could've sent a pull request for this :)

joshmedeski:
We'll have to see what @lra thinks of this change. I'm comfortable with making a pull request but I want to see if he approves of this change before I take the time to work on it.

GentleHoneyLover:
@joshmedeski, @Ira, is there an appetite to work on this further?

yamatsum:
@joshmedeski 
@lra seems to agree to set up mackup with XDG Base Directory.
https://github.com/lra/mackup/issues/1317#issuecomment-460076807

tshu-w:
vote for this

PR #1695, opened 2021-02-13
Update neovim to support init.lua

olimorris:
Neovim from v0.5 onwards will support an init.lua config file (as per this [commit](https://github.com/neovim/neovim/commit/72c22862dc2199462aef0d450a49d29a9d0680b9)). This pull request adds support for this.

### DIFF
--- a/mackup/applications/neovim.cfg
+++ b/mackup/applications/neovim.cfg
@@ -7,3 +7,4 @@ name = neovim
 
 [xdg_configuration_files]
 nvim/init.vim
+nvim/init.lua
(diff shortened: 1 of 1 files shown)


yamatsum:
I think have to add `lua/` directory as well

olimorris:
Good point. Common practice will be to have modules loaded from the lua directory in the `init.lua` file.

alexhumphreys:
Was just about to open the same PR 😅  Any reason this can't be merged?

lra (approved):

PR #1707, opened 2021-03-12
add support for CleanShot

ye4241:
ref https://cleanshot.com/


lra:
You have a bunch so I'll do it, thanks.

lra (approved):

PR #1716, opened 2021-03-14
Add support for Logitech Options

damianoamatruda:
Logitech Options: https://www.logitech.com/en-us/product/options


joshmedeski:
Yay, thanks for adding this 👍

Issue #1727, opened 2021-04-11
French spacing


LucasLarson:
It looks like some of this was built by a français with French spacing` `? We should remove the spaces between sentences and their question marks and exclamation points like this:
```diff
-Are you sure that you want to replace it ?
+Are you sure that you want to replace it?
```
Example taken from: https://github.com/lra/mackup/blob/9f275ed1589c55273eac4a71ff3bf8c480860c15/mackup/application.py#L104-L105

lra:
You are right 🇫🇷

LucasLarson:
resolved in https://github.com/lra/mackup/commit/442a4add4c9955c713655cc2520a876b5ac7a385

PR #1728, opened 2021-04-11
repair punctuation spacing (#1727)

LucasLarson:
- [x] move punctuation from `fr-FR` to `en-US`

### DIFF
--- a/mackup/mackup.py
+++ b/mackup/mackup.py
@@ -36,7 +36,7 @@ class Mackup(object):
                 " Run mackup --help for guidance."
             )
 
-        # Do we have a folder to put Mackup into?
+        # Do we have a folder set to save Mackup content into?
         if not os.path.isdir(self._config.path):
             utils.error(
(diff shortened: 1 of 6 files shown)


mackup/mackup.py
```diff
@@ -36,13 +36,13 @@ def check_for_usable_environment(self):
                 " Run mackup --help for guidance."
             )
 
-        # Do we have a folder to put the Mackup folder ?
+        # Do we have a folder to put Mackup into?
```
lra, line 39: It's really not Mackup that we put, but the Mackup data files. Can you reword to make it more explicit?

lra (commented):
Thanks, one small comment.

mackup/mackup.py
LucasLarson: > make it more explicit?

You’re right – let me know if c638fd8 doesn’t go far enough.

lra (approved):

PR #1732, opened 2021-04-23
Add support for Sublime Text 4

TCattd:
Add support for Sublime Text 4. Or Sublime Text. No number at the end.

Settings folder now has no version number, and un-official announcement (Discord https://tppr.me/gkDTU ) says that Sublime Text will drop version numbers entirely (will keep his internal build numbers, of course).

Builds from 4000 branch (available in official Sublime's Discord) use this new path without number on it to save settings.


lra (approved):
Thx!

PR #1754, opened 2021-08-04
Added support for Affinity

EpicKris:


mackup/applications/affinity-designer.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = Affinity Designer
+
+[configuration_files]
+Library/Containers/Affinity Designer/Data/Library/Application Support/user
```
lra, line 5: All files need to end with a newline character, I'll fix it in those 3 files.

lra (approved):
Thanks!

Issue #1778, opened 2021-11-22
Feature Request: Option to use hardlink for specific apps


dvcrn:
Some apps aren't happy with softlinks and overwrite them right away. Would be nice to have a hardlink flag to create a hardlink instead of a softlink for these files

PR #1779, opened 2021-11-22
Add support for hardlinkink files

dvcrn:
## What

Refs https://github.com/lra/mackup/issues/1778

Adds support for specifying `mode` in application config files (currently `softlink` or `hardlink`, uses `softlink` when not set)

When `hardlink` mode is used, will use `os.link` instead of `os.symlink` to create a hardlink for files instead

This was a quick hack looking at this codebase for the first time, so this is by no means ready to go. I wanted to submit a proof of concept PR anyway to see what the general opinion on this is. 

Caveats: 
- Doesn't work with directories on mac, so I disabled hardlink when a directory is used as target, and defaulted back to softlink
- Also won't work when the file is on a different partition/drives so we need to add a check for this

## Why

There are a few applications that don't play nicely with softlinks and just override them on start (bartender, alt-tab, etc). Using a hardlink instead of a softlink goes around this issue because the app (usually) doesn't check the link count of config files. For it, the config file is just the same usual file on disk. 

### All submissions

* [ ] My submission passes the [tests](https://github.com/lra/mackup/tree/master/tests)
* [x] I have linted the code locally prior to submission
* [ ] I have written new tests as applicable
* [x] I have added an explanation of what the changes do


lra:
The problem with hardlinks is, like you mentioned, that 1. macos does not like them 2. lots of files hosting services break when used (e.g. dropbox).

So it's hard to merge such a change =/

dvcrn:
Actually mac supports hardlinks just fine (through the normal `ln a.txt b.txt` command). The only thing it doesn't support are hardlinks to folders, that got axed in APFS and was only usable through hacks previously. But hardlinks to folders aren't that advisable anyway so they won't be missed 😄 

For Dropbox - Dropbox like other apps treats hardlinks just like normal files and syncs them fine. I'm using hardlink for a couple different things and didn't have issues yet. 

What Dropbox can't do is having hardlinks in dropbox to other files within dropbox. It ignores the link count completely so file1 and file2 will both be synced as completely separate files with identical content

lra:
> But hardlinks to folders aren't that advisable anyway so they won't be missed

Lots of Mackup apps do link folders.

> What Dropbox can't do is having hardlinks in dropbox to other files within dropbox. It ignores the link count completely so file1 and file2 will both be synced as completely separate files with identical content (syncing 2 hardlinked files down from dropbox results in 2 files with identical content without any link)

I see, when I tested this, the behavior was quite inconsistent, I guess I need to check it out again. But we'd need to test it with other engines too, and OSes.

dvcrn:
> Lots of Mackup apps do link folders.

Ah that's what you meant, yes I agree! This was more thought to be a solution for single config files that get overwritten by the target application on launch

Folders could still be just symlinks like usual (how I currently implemented it in this PR as well). 

> Also, instead of having a "sync mode" per app, what about adding a section to describe hardlink files instead (e.g. "hardlink_configuration_files" ? This way an application config could support both. I'm just thinking this might be an edge case for some app in the future. Or maybe it will just make everything more confusing, no strong opinion here.

I like this idea better! It makes this solution something for edge cases without changing existing behavior. I'll try a quick implementation and see how it goes

But yes we'll have to properly test this against different engines. Personally only using Dropbox and iCloud... 💭

dvcrn:
Actually turns out my plan didn't work as I hoped it would so this might not be feasible after all 😢 

iCloud at least overwrites the file completely when a change happens instead of writing it with new data, so whenever iCloud syncs down, it breaks the hardlink. Too bad. 

I wonder what else can be done to fix apps that don't play nicely with symlinks

lra:
Yes, a big caveat of Mackup is that it does not have its own sync engine, so it has to assume users use the worst of the sync engines available =( And starting to make edge cases based on sync engines is a nightmare.

Thanks for trying though!

PR #1786, opened 2021-12-18
Add support for .vim/bundle folder.

iarchean:
since my all vim plugins are placed on .vim/bundle folder, so add support for it.


lra (approved):
Thanks!

PR #1814, opened 2022-03-26
Fix code formatting and "Lint Python" GitHub Action

dnicolson:
This pull request updates the `black` target version to Python 3.10 as Python 2 support was dropped in https://github.com/psf/black/releases/tag/22.1.0. This should also fix the GitHub Action that is failing.


lra (approved):

lra:
thanks!

PR #1847, opened 2022-08-09
Add support for VSCode Extensions

nyshk97:
I have read discussions below.
https://github.com/lra/mackup/pull/1510
https://github.com/lra/mackup/pull/1170

Still I think it's useful to sync VSCode extensions, and I guess not a few people think the same.

Considering the number of VSCode users, it's better to sync VSCode extensions out of the box and **disable only the extension part** if you want, than force to investigate where extensions settings are and create `cfg` file from scratch.

### All submissions

### Adding/updating Application X Support

* [x] This PR is only for one application
* [x] It has been added to the list of supported applications in the [README](https://github.com/lra/mackup/blob/master/README.md)
* [x] Changes have been added to the WIP section of the [CHANGELOG](https://github.com/lra/mackup/blob/master/CHANGELOG.md)
* [x] Syncing does not break the application
* [x] Syncing does not compete with any syncing functionality internal to the application
* [ ] The configuration syncs the minimal set of data
* [x] No file specific to the local workstation is synced
* [x] No sensitive data is synced

### DIFF
--- a/mackup/applications/vscode-extensions.cfg
+++ b/mackup/applications/vscode-extensions.cfg
@@ -1,0 +1,5 @@
+[application]
+name = name = Visual Studio Code Extensions
+
+[configuration_files]
+.vscode/extensions
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Updated support for Proxymann Setapp version (via @JanC)
 - Added support for Btop (via @Mersid)
 - Updated support for Nushell (via @maradude)
+- Added support for VSCode Extensions (via @nyshk97)
 
 ## Mackup 0.8.34
 
--- a/README.md
+++ b/README.md
@@ -672,6 +672,7 @@ See the [README](doc/README.md) file in the doc directory for more info.
 - [Vimperator](http://www.vimperator.org/vimperator)
 - [Viscosity](http://www.sparklabs.com/viscosity/)
 - [Visual Studio Code](https://code.visualstudio.com/)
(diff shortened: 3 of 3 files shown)


mackup/applications/vscode-extensions.cfg
```diff
@@ -0,0 +1,5 @@
+[application]
+name = name = Visual Studio Code Extensions
```
lra, line 2:
```suggestion
name = Visual Studio Code Extensions
```

lra (commented):
I don't know what to do here, I don't like the idea of starting splitting applications into multiple configs, it will make everything more complex and confuse people, which will lead to issues to answers.

It was also already added and removed.

Why not using https://github.com/lra/mackup/tree/master/doc#add-support-for-an-application-or-almost-any-file-or-directory ?
What about having

nyshk97:
@lra 
OK, I understand your thoughts.

Probably you are right.
I've just thought my personal setup would be useful to someone else, but it certainly complicates things, as you say.

Sorry to have confused you.

Thanks for the great application.
I will keep using this.

lra:
Thanks, feel free to close it.

> Also, it seems you can sync extensions officially: https://code.visualstudio.com/docs/editor/settings-sync

What's wrong with this? Why do you still need to use Mackup for vscode?

nyshk97:
> What's wrong with this? Why do you still need to use Mackup for vscode?

I've just felt that it's useful to manage with fewer tools.

I will close this issue. Thanks anyway.

PR #1908, opened 2023-02-20
Add support for TablePlus Favorites

pesonn:
Add support to backup query favorites

https://docs.tableplus.com/query-editor/query-history#save-to-favorite


lra (approved):
Thanks

PR #1987, opened 2024-01-19
add tests and changelog

NorthIsUp:


con-f-use:
How has that not been merged yet?

con-f-use:
pretty plz.

lougreenwood:
Seems like a pretty straight forward & sensible PR for a long-requested feature. @lra any thoughts on this PR? Thanks! 🙏

coquizen:
Agreed, this should be implemented. Those concerned with their config settings are often concerned with organizing where those config files are stored.

PR #2012, opened 2024-04-07
Add option --copy-only

dracorp:
### All submissions

### Adding/updating Application X Support

* [x] This PR is only for one application
* [x] It has been added to the list of supported applications in the [README](https://github.com/lra/mackup/blob/master/README.md)
* [x] Changes have been added to the WIP section of the [CHANGELOG](https://github.com/lra/mackup/blob/master/CHANGELOG.md)
* [x] Syncing does not break the application
* [x] Syncing does not compete with any syncing functionality internal to the application
* [x] The configuration syncs the minimal set of data
* [x] No file specific to the local workstation is synced
* [x] No sensitive data is synced

### Improving the Mackup codebase

* [x] My submission passes the [tests](https://github.com/lra/mackup/tree/master/tests)
* [x] I have linted the code locally prior to submission
* [ ] I have written new tests as applicable
* [x] I have added an explanation of what the changes do

### DIFF
--- a/mackup/main.py
+++ b/mackup/main.py
@@ -17,6 +17,7 @@ Options:
   -f --force    Force every question asked to be answered with "Yes".
   -r --root     Allow mackup to be run as superuser.
   -n --dry-run  Show steps without executing.
+  -c --copy-only   Copy only files when backup.
   -v --verbose  Show additional details.
   --version     Show version.
 
@@ -36,13 +37,15 @@ See https://github.com/lra/mackup/tree/master/doc for more information.
 
 """
 
+import sys
+
 from docopt import docopt
-from .appsdb import ApplicationsDatabase
+
+from . import utils
 from .application import ApplicationProfile
+from .appsdb import ApplicationsDatabase
 from .constants import MACKUP_APP_NAME, VERSION
 from .mackup import Mackup
-from . import utils
-import sys
 
 
 class ColorFormatCodes:
@@ -83,13 +86,17 @@ def main():
 
     verbose = args["--verbose"]
 
+    copy_only = args["--copy-only"]
+
     if args["backup"]:
         # Check the env where the command is being run
         mckp.check_for_usable_backup_env()
 
         # Backup each application
         for app_name in sorted(mckp.get_apps_to_backup()):
-            app = ApplicationProfile(mckp, app_db.get_files(app_name), dry_run, verbose)
+            app = ApplicationProfile(
+                mckp, app_db.get_files(app_name), dry_run, verbose, copy_only
+            )
             printAppHeader(app_name)
             app.backup()
 
@@ -100,7 +107,8 @@ def main():
         # Restore the Mackup config before any other config, as we might need
         # it to know about custom settings
         mackup_app = ApplicationProfile(
-            mckp, app_db.get_files(MACKUP_APP_NAME), dry_run, verbose
+            mckp, app_db.get_files(MACKUP_APP_NAME), dry_run, verbose,
+            copy_only
         )
         printAppHeader(MACKUP_APP_NAME)
         mackup_app.restore()
@@ -116,7 +124,9 @@ def main():
         app_names.discard(MACKUP_APP_NAME)
 
         for app_name in sorted(app_names):
-            app = ApplicationProfile(mckp, app_db.get_files(app_name), dry_run, verbose)
+            app = ApplicationProfile(
+                mckp, app_db.get_files(app_name), dry_run, verbose, copy_only
+            )
             printAppHeader(app_name)
             app.restore()
 
--- a/mackup/main.py
+++ b/mackup/main.py
@@ -13,13 +13,13 @@ Usage:
   mackup --version
 
 Options:
-  -h --help     Show this screen.
-  -f --force    Force every question asked to be answered with "Yes".
-  -r --root     Allow mackup to be run as superuser.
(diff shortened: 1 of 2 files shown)


huyz:
I find the name `copy-only` hard to understand.  I had to look at the code changes to understand. I think `preserve-source` might be clearer.

mackup/main.py
```diff
@@ -13,12 +13,13 @@
...
-  -f --force    Force every question asked to be answered with "Yes".
-  -r --root     Allow mackup to be run as superuser.
-  -n --dry-run  Show steps without executing.
-  -v --verbose  Show additional details.
-  --version     Show version.
+  -h --help         Show this screen.
+  -f --force        Force every question asked to be answered with "Yes".
+  -r --root         Allow mackup to be run as superuser.
+  -n --dry-run      Show steps without executing.
+  -c --copy-only    Copy only files when backup.
```
huyz, line 20: Idiomatic English would be more like: "Only copy files when backing up"

huyz:
I feel like this behavior should be the default.  Unless I'm mistaken, the idea of symbolic links doesn't really work for many apps, e.g. Preferences file. As soon as an app rewrites the Preferences file, the symbolic link is broken.

I figure the default model should be copy/sync, possibly merge, but not symlinking, unless we can be assured that symlinks don't get deleted.

Am I understanding mackup right? I'm new to this app.

martimlobao:
Would it be possible to enable this per-application as well? I just realized that iStat Menus also breaks using the symlinks created by Mackup, and so it would nice to be able to modify a `.cfg` file with a flag to have it only copy files instead.

dracorp:
> I feel like this behavior should be the default. Unless I'm mistaken, the idea of symbolic links doesn't really work for many apps, e.g. Preferences file. As soon as an app rewrites the Preferences file, the symbolic link is broken.
> 
> I figure the default model should be copy/sync, possibly merge, but not symlinking, unless we can be assured that symlinks don't get deleted.
> 
> Am I understanding mackup right? I'm new to this app.

So, I've added a few apps to ignore, vim, iterm2 etc.

martimlobao:
> You mean make it part of the config file for each app? That sounds like a good idea.

Yeah, unless I'm reading it wrong, this PR currently only provides a CLI flag, so it's all or nothing. Having the granularity to configure which apps use the copy/paste method instead of symlinks will probably address a lot of the issues some apps appear to be having with symlinks.

huyz:
> So, I've added a few apps to ignore, vim, iterm2 etc.

@dracorp What do you mean there? I don't see any changes in the code that would do what @martimlobao is asking for.

dracorp:
> @dracorp What do you mean there? I don't see any changes in the code that would do what @martimlobao is asking for.
It's not up to me anymore. I only cared about copying. 

> > So, I've added a few apps to ignore, vim, iterm2 etc.
This is in my mackup configuration.

victor-falcon:
IMO, the copy option should be by default, because right now Mackup is not working properly with a lot of apps :(. Maybe we can do that with a global flag in the config file if we don't want to modify the default behavior.

dracorp:
Is there any chance of someone completing this? I'm not that proficient in Python to impose anything.

kzaoaai:
This still doesn't do --copy-only for restoring. I am no expert but the following modification to def restore inside application.py does it for me:

```
    def restore(self):
....
....
                    if utils.confirm(
                        "You already have a {} named {} in your"
                        " home.\nDo you want to replace it with"
                        " your backup?".format(file_type, filename)
                    ):
##NEW CODE START##############################################################
                        # For --copy-only option, just copy the file and skip the rest
                        if self.copy_only:
                            # Copy the file
                            utils.copy(mackup_filepath, home_filepath)
                            continue
##NEW CODE END################################################################
                        utils.delete(home_filepath)
                        utils.link(mackup_filepath, home_filepath)
                else:
##NEW CODE START##############################################################
                    # For --copy-only option, just copy the file and skip the rest
                    if self.copy_only:
                        # Copy the file
                        utils.copy(mackup_filepath, home_filepath)
                        continue
##NEW CODE END################################################################
                    utils.link(mackup_filepath, home_filepath)
            elif self.verbose:
```

bam80:
@lra ping. It was pushed a year ago, we need your response.

SmetDenis:
Waiting for it so bad!

sedinamwright330:
Ok

null

kennyparsons:
yes, I'd love to see this come through. Apps break with symlinks. I just want to run mackup backup and have --copy-only be default, so i can point in time snapshot everything (and subsequently git commit those changes, but that's outside the purview of this app).

Kol9yN:
@lra Collegue, please, pay attention to this ticket if you can, here very good and important solution

evanharmon1:
It really sounds like a copy-only option will solve a lot of things with the tool. Any chances there is movement on this? I would love to keep using this awesome tool but at this point it keeps causing issues and breaking things.

shaunakv1:
Any update on this? this is badly needed!  Symlinks don't work for iterm2 for me anymore and it refuses to save config?

SmetDenis:
120% the same issue and different tool. Please! Merge it.

lra:
Thanks, this has been implemented in https://github.com/lra/mackup/releases/tag/0.9.0

driesvints:
thanks for your work @lra

Issue #1854, opened 2022-09-01
Symlinks replaced by applications (for Library/Preferences)


jonasdiemer:
Hi,

I noticed that applications like BetterTouchTool and iTerm2 replace the symlinks created by mackup in `~/Library/Preferences/` with physical copies of the config files.

Is this expected? Would `mackup restore` still allow me to sync preferences to a device? 

Regards
Jonas

azuenko:
For anyone facing the same issue with iTerm2, a quick workaround would be to say iTerm2 to store the configuration in a custom folder, e.g. `~/.config/iterm2`. Then, use a custom `iterm2-ext-config.cfg` :

```
[application]
name = iTerm2-ext-config

[xdg_configuration_files]
iterm2
```

Unfortunately, this approach works only for iTerm, and if an app keeps replacing a symbolic link with a file in, and there's no way to isolate that file in a separate folder, I'm afraid there's nothing can be done here.

jonasdiemer:
Found some discussions on how to approach it here: https://apple.stackexchange.com/questions/385102/how-to-properly-symlink-files-in-the-library-preferences

One option might be to lock the file (`chflags nouchg`).

jonasdiemer:
This behavior has nothing to do with the storage destination (iCloud or whatever), as the symlink that is overwritten lies on the "source storage".

jonasdiemer:
No, the symlinks that are put in place of the plist files have nothing to do with the storage destination - they simply point to another path in the filesystem (AFAIK).

lra:
> Is it fair to assume this might lead to Mackup being obsolete?

If every application and OS decide to throw away the POSIX concept of file linking, yes. But before it happens, the strategy has always been to remove support for those impacted applications in Mackup.

anthonysomerset:
This appears to me to be specific to MacOS Ventura (13) - I don't appear to observe this issue on older releases

Morriz:
> This appears to me to be specific to MacOS Ventura (13) - I don't appear to observe this issue on older releases

If it would be as easy as adding that `chflags nouchg` operation, then I hope it will be added soon :)

Triloworld:
Owner of Raycast start to remove sync options without anybody notice because of this issue: https://github.com/lra/mackup/issues/1980#issuecomment-1884729862
@gridsystem have right. Applications may have they own solution that block copy operation but it seams that is system issue

Issue #2008, opened 2024-03-26
Add an option --copy-only to copy only files to mackup folder


dracorp:
Can you add an option **--copy-only** which copies only files to mackup folder? Similar to **--dry-run** but doing a little more.

Sometime I want just copy/backup files, nothing more.

dracorp:
https://github.com/lra/mackup/pull/2012

huyz:
This really should be the default. So many apps do not expect symbolic links and re-create the files anyway.

lra:
Implemented in https://github.com/lra/mackup/releases/tag/0.9.0

PR #2072, opened 2025-07-06
feat: add option to copy instead of link on app level

kassi:
This pull request adds an application specific setting `symlink = false|true` to application config files.
If this setting is set to `false`, then no files will be linked, but instead copied over.

In this version, the default value, if the setting does not exist, is `true` to not break backwards compatibility.

May close: #1854
May obsolete: #2008 

**Scope** 
This version allows for copying **all** files of one application instead of linking them.
It's not meant to work for individual files only, like link _filea_ but do not link _fileb_. This would add another level of complexity, but also of complication inside the config files, either by adding a tag to each file to sync or have different sections like `configuration_files_to_link` and `configuration_files_to_copy` and the same for `xdg_configuration_files`.
In order to keep it clean and simple, I chose the provided variant.

### All submissions

* [X] I have followed the [Contributing Guidelines](https://github.com/lra/mackup/blob/master/.github/CONTRIBUTING.md)
* [X] I have checked to ensure there aren't other open [Pull Requests](https://github.com/lra/mackup/pulls) for the same update/change

### Improving the Mackup codebase

* [X] My submission passes the [tests](https://github.com/lra/mackup/tree/master/tests)
* [X] I have linted the code locally prior to submission
* [X] I have written new tests as applicable
* [X] I have added an explanation of what the changes do

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -157,6 +157,23 @@ in your home folder:
 cp mackup/doc/.mackup.cfg ~/
 ```
 
+### Sync an application that doesn't work with symbolic links
+
+In the application specific configuration, either as part of this repository or in your home folder's `.mackup` folder,
+add the option to **not** symlink:
+
+```ini
+[application]
+name = Alfred
+symlink = false
+
+[configuration_files]
+Library/Preferences/com.runningwithcrayons.Alfred-Preferences.plist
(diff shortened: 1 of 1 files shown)


lra:
Thanks, this has been implemented in https://github.com/lra/mackup/releases/tag/0.9.0

PR #2085, opened 2025-10-15
Implement Copy Mode and Refactor Backup/Restore Operations

lra:
Adds "copy mode" as an alternative to "link mode" for managing app settings, enhancing flexibility. Refactors backup and restore operations to support both modes, renames methods (e.g., `backup()` to `link_install()`), and updates README/docs.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -19,7 +19,8 @@ mackup -h
 
 ### With Homebrew (OSX only) master branch for latest updates
 
-Want to install the latest master release instead of waiting on the homebrew package version?
+Want to install the latest master release instead of waiting on the homebrew
+package version?
 
 [Homebrew reference](https://docs.brew.sh/Manpage#install-options-formulacask)
 
--- a/doc/configuration_merge_guide.md
+++ b/doc/configuration_merge_guide.md
@@ -78,7 +78,7 @@ across the machines. We decide that Machine A will serve as our master.
 First, backup the bash configuration files (there are a few of them)
 for your application on all machines.
 
-##### Sample backup commands for Machine A
+##### Method 1: Sample backup commands for Machine A
 
 ```bash
 mkdir ~/bash_backup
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -108,49 +95,73 @@ Get some help, obviously...
...
+which can be in Dropbox, iCloud, or wherever you configure it.
+
+It is covered by the 2 commands:
+
+- `mackup backup`
+- `mackup restore`
+
+### Link mode
+
+Link mode is used to move your config files into the Mackup folder,
```
devnoname120: What would you think about adding a warning about the fact that the link method is dangerous on macOS Sonoma and later?

mackup/application.py
```diff
@@ -136,9 +242,9 @@ def backup(self) -> None:
                 else:
                     print("Doing nothing\n  {}\n  does not exist".format(home_filepath))
 
-    def restore(self) -> None:
+    def link(self) -> None:
```
devnoname120: And maybe add a runtime warning or a confirmation dialog too on macOS Sonoma and later

mackup/main.py
```diff
@@ -21,12 +22,13 @@
...
- 2. backup: sync your conf files to your synced storage, use this the 1st time
-    you use Mackup.
- 3. restore: link the conf files already in your synced storage on your system,
-    use it on any new system you use.
- 4. uninstall: reset everything as it was before using Mackup.
+ - mackup list: display a list of all supported applications.
+ - mackup show: display the details for a supported application.
+ - mackup backup: copy local config files in the configured remote folder.
+ - mackup restore: copy config files from the configured remote folder locally.
+ - mackup link install: moves local config files in remote folder, and links them.
```
devnoname120: Ditto

lra: https://github.com/lra/mackup/pull/2085#discussion_r2462653731

README.md
devnoname120: My bad I somehow missed it

mackup/application.py
lra: Yes, it's super annoying, and my guess is that it's related to what the terminal running mackup can do... 🤷‍♂️

